PR #23, opened 2024-09-03
Adds floating point sample functions for covergroup generation

ross144:
I had to split the floating point convert and move instructions into multiple types. The sample functions need to distinguish between rd and fd and rs1 and fs1 and there are three variations.  Float to float (looks just line floating point I type), float to  integer, and integer to float.

@davidharrishmc can you check this is to your liking?


davidharrishmc:
I got Zfh.

PR #37, opened 2024-09-05
updated makefile to create dirs if they don't exist

AnonymousVikram:
Ran into an error when running `make` and `make sim` where it would error out when the directories don't exist


davidharrishmc:
This is great to be improving the code already.  The SRCDIR mkdir seems out of place  in the middle of a batch of defining variables.  Could you move it to a more logical location?

davidharrishmc:
Nice work contributing to the code base already.

PR #53, opened 2024-09-10
Fixed bugs in templates and testplans preventing wsim from running.

davidharrishmc:


ross144:
@davidharrishmc Can you add the SPDX header to fix.py?

PR #58, opened 2024-09-11
Fixed Issue#31

Zain2050:
Fixed issue#31 in addi, now bins with negative values are being hit. The tests for cr_rs1_imm_corners were not being generated due to an issue in its template. It is also fixed.

The coverage of addi is now 97.33% which initially was 75.06%.


davidharrishmc:
Great fix.  Can you remove the Makefile from this PR, because each team will want their own Makefile sim emphasizing the instructions that team is working on this week?  Just post the addi fix.

PR #59, opened 2024-09-12
100% coverage of lw achieved

Mysterio-Abdullah:
Excluded x0 registers from cp_rs1 coverpoint, also removed cp_mem_unaligned coverpoint and got 100% coverage of cp_rd_corners.


Huda-10xe:
Great work. Could you remove the Makefile from this PR, and just contribute the coverage fixes, as Makefile will vary for every contributor depending on the instruction they're working on.

davidharrishmc:
In line 197 of testgen.py, should the second rs2val be immval?
In line 205, should the first v be -1?  The comment at 202 says rs1 = all 1s, rs2 = v.
cp_rs1_nx0 does not list a bin for x0.  It also lists an ignore_bins x0_bin.  Is it necessary to have the ignore_bins if there isn’t a bin listed?

PR #61, opened 2024-09-12
100% coverage for auipc

ahmad101-sudo:
Changes:
Added a new template file named "RV32_cp_rd_corners_auipc.txt"
Made changes in testgen.py
Added a new coverpoint in RV32I testplan


davidharrishmc:
make_rd_corners_auipc calls writeCovVector three times.  Are all three necessary, or is 1 sufficient for this purpose?  Take a look at the test cases you are generating in the .S and I think you'll be able to streamline this one.

The more general make_rd_corners has many cases so it will work with many different instructions.

davidharrishmc:
Sorry to be picky here, but the rs1val and rs2val in writeCovVector is presently set to v, along with the immVal.  In case there was some weird bug that caused auipc to be influenced by rs1 or rs2, it would be better for those values not to match the immediate.  Can you use the random values rs1val and rs2val instead?

Also, please remove the Makefile from the PR.  Each of us is going to have our own Makefiles to test our own instructions, and there's no need to keep changing them in the repository - it's fine to have your own local copy instead.

davidharrishmc:
Now there's a merge conflict.  Also there still seem to be v, v, v  in writeCovVector

ahmad101-sudo:
merge conflict has been solved.

PR #65, opened 2024-09-13
100% coverage of RV32 addi achieved

Zain2050:
Modified covergroupgen.py to generate a new block for 'mv' instruction in covergroup.svh file.


davidharrishmc:
I like this solution.  It will be hard to understand the reason for the special case a year from now from the comments in the file.  Could you flush out the comment to explain why this is necessary, in a way somebody unfamiliar with the issue will appreciate quickly.

Zain2050:
I made the requested change. Hope its better.

PR #66, opened 2024-09-14
updated makefile to enable other extensions

AnonymousVikram:
Other extensions letters should be added in line `42` of `Makefile` (comment exists to guide user)


jordancarlin:
Might be nice to move this (and for that matter all of the variables) up to the top of the Makefile. It’ll be easier to find and is more in line with standard conventions for Makefiles.

davidharrishmc:
@AnonymousVikram could you take another look at this?  We still have to manually add extensions in the Makefile this way.  It would be cleaner for the Makefile to just search all of the subdirectories in rv32 and rv64 and build the code in them, rather than manually specifying which directories are interesting.

Also, I don't think we need the mkdir -p to create such directories.  The .S files will exist in certain directories before we run make build.  So the real problem is to identify which directories exist, and build code in all of them, rather than to create empty directories that don't already exist.

Makefile
```diff
@@ -34,11 +34,11 @@ SRCEXT 		:= \([$(CEXT)$(AEXT)$(SEXT)]\|$(CPPEXT)\)
 #SRCS = $(wildcard *.S)
 #PROGS = $(patsubst %.S,%,$(SRCS))
 BASEDIR = ${WALLY}/addins/cvw-arch-verif
-SRCDIR64 = ${BASEDIR}/tests/rv64/I
-SRCDIR32 = ${BASEDIR}/tests/rv32/I
+SRCDIR64 = ${BASEDIR}/tests/rv64
+SRCDIR32 = ${BASEDIR}/tests/rv32
+WALLYEXTS = 	$(shell find ${SRCDIR64} ${SRCDIR32} -type d | sort)
```
jordancarlin, line 39: This variable is no longer doing anything

AnonymousVikram: @jordancarlin I left the variable there for extensibility, in case the folder structure needs to be changed. It also makes it more readable

jordancarlin: Sounds good. Just double checking.

PR #71, opened 2024-09-14
Coverage of ld achieved.

Mysterio-Abdullah:
Issues were similar to lw, but for cp_rd_corners, some changes were required in testgen.py.


davidharrishmc:
@Mysterio-Abdullah there are merge conflicts.  It looks like you are trying to push changes like the make_rd_corners_auipc and other pieces of code from testgen.py that are already in the repo.  Please do a "git pull usptream main" and fix any merge conflicts locally before you push.  It's also a good idea to pull upstream before you start any new work, so you are less likely to introduce conflicts.

PR #73, opened 2024-09-15
lh RV32I, complete coverage.

Mysterio-Abdullah:
Created a new template for 16 bits half word. Added 16 bits corners in testgen.py and write a function to make it cp_rd_corners for 16 bits. Also already excluded x0 in previous commits. If anything seems strange, please let me know.


davidharrishmc:
make_rd_corners_16 can be more like make_rd_corners_auipc in that it only needs one writeCovVector with the expect value; the sources don't have to be twiddled.  Moreover, they only differ in the set of corners you are using.  How about having a single make_rd_corners_1(test,xlen, corners) that you pass either corners_16bits (for lh), corners_16ubits (for leu), corners (for auipc), and various other corner cases for lb, lbu, etc?

Mysterio-Abdullah:
Yeah, sounds good. I'll try and send another PR soon.

Mysterio-Abdullah:
Sir, please have a look at new commit. I think it's better.

PR #77, opened 2024-09-18
RV64 auipc 100% coverage

ahmad101-sudo:
added a new template
checked new coverpoint in RV64I.csv
In case of RV64 instructions "make sim" was giving error messages so i uncommented line 98 from covergroupgen.py.


davidharrishmc:
PR #955 on cvw fixes the path of the rv64gc coverage files so you don't need to uncomment covergroupgen.  Give one of the committers a little bit to accept 955, then try this again without the covergroupgen change.

Sorry for the delay on 955!

jordancarlin:
@ahmad101-sudo PR openhwgroup/cvw#955 is merged so this should work now

ahmad101-sudo:
@jordancarlin Thanks it worked.
@davidharrishmc I have committed again without the changes in covergroupgen.py

PR #87, opened 2024-09-20
lh RV64I 100% Coverage

Mysterio-Abdullah:
covergen.py was reading the .csv file but couldn't distinguish between RV32_cp_rd_corners_lh.txt & RV64_cp_rd_corners_lh.txt and writing the coverage of RV32 instead of RV64. That's why I made some changes in the new template and named it RV64_cp_rd_corners_lh64.txt. Now, the coverage for lh RV64I has been achieved. If anything can be improved, please let me know.


davidharrishmc:
That shouldn't be necessary and looks like a bug that will affect everyone else as they need RV32 vs. RV64.

Can you look further to fix the root cause rather than hacking around this?

Mysterio-Abdullah:
Ok Sure, I'll look into it. And make another PR soon.

PR #88, opened 2024-09-20
lh RV64I 100% Coverage

Mysterio-Abdullah:
There was a priority issue in a conditional statement of covergen.py. The elif (prefixName in covergroupTemplates) should come before the generic if (name in covergroupTemplates) check i beleive. The issue has been resolved now and coverage of lh RV64I has been completely achieved without any hack :)
Thanks for pointing :)


davidharrishmc:
That's a fine fix to reverse the order.  

It looks like the root cause is that there are now three flavors of this template:
cp_rd_corners_lh.txt
RV32_cp_rd_corners_lh.txt
RV64_cp_rd_corners_lh.txt

When you add the RV32 or RV64 flavor, delete the one with no prefix because it should not longer be used.  That will also solve the issue.  Can you update the PR once more?  Let me know if it seems I'm overlooking something.

Mysterio-Abdullah:
Yup, Sir, you're right cp_rd_corners_lh.txt is of no use now. Deleting it can also reslove the issue.

PR #94, opened 2024-09-23
Starting to add compressed and floating-point coverage, but 0% so far

davidharrishmc:


davidharrishmc:
Now has 38% coverage on fadd.s.  The flow is working for RV32F.

PR #96, opened 2024-09-24
Majority coverage of 64 & 32 bit fsw

coreyqh:


davidharrishmc:
Doesn’t fsw take one fpr source and one gpr base address instead of two fpr
sources?

Please don’t check in the make file. All of us are tweaking it for our
individual tests.

coreyqh:
Sorry about checking in the Makefile, I'll try to be more careful. 

Yes, you are correct that fsw takes one int and one float argument. I simply added the cp_fs1, cp_fs2, cp_fs1_corners, and cp_fs2_corners coverpoints since they were missing, and since I needed two of them why not make all four and help everyone else out in the process.

I believe that I all of the tests and coverpoints reflected the correct register arguments, unless there is something else you see that I am missing? 

In either case I am going to close this PR and submit another one later, hopefully with full coverage (and definitely without the Makefile).

PR #97, opened 2024-09-24
Full coverage of fsw in RV32, near full coverage in RV64

coreyqh:
Still sorting out fs2_corners issues keeping fsw from full coverage in RV64

While fsw only requires cp_fs2 and cp_fs2_corners, I implemented cp_fs1 and cp_fs1_corners as well since there are essentially identical.


davidharrishmc:
The fsw is writing to the same address it just loaded from.  If store didn’t work, we wouldn’t know because the memory address already had the value we are trying to store.  I think you want to introduce a second scratch address, maybe 8 bytes past the current scratch address.  Perhaps you can add 8 to signedImm12 to get there without needing another register.

PR #100, opened 2024-09-25
Partial coverage for fmadd.s, support for other FR4-type instructions

HamzaYusufJamal:
cp_fpr_hazard for fmadd.s hangs the sim due to a bug in the assembly tests. For now, to get a functioning simulation, cp_fpr_hazard is not checked for fmadd.s, but it should be added back in later once the bug is resolved.


davidharrishmc:
@HamzaYusufJamal there was an if rather than elif that was causing errors to be reported on r-type instructions during testgen.  PR 102 fixes this.

PR #101, opened 2024-09-26
Modified testgen to enable fp once at the top of assembly files

coreyqh:


coreyqh:
Recent commit ensured all currently implemented floating point instruction types have mstatus.FP set to enable floating point at the top of their respective assembly files.

PR #104, opened 2024-09-26
slt RV32I 100% Coverage

TheUnparalleledDragon:


davidharrishmc:
There is already a  RV32_cp_rd_corners_lui.txt template for lui.  It is used automatically for RV32, and we will eventually need an RV64_cp_rd_corners_lui.txt template.  The cp_rd_corners_lui.txt is redundant and will just lead to trouble if they get inconsistent.  Please remove it from the PR; let's discuss if that causes you any problems.

PR #107, opened 2024-09-26
Made the coverage for lui 64 to go upto 40 percent, me and ehzem will…

Abdullah-Mujtaba:
… work on it more tommorrow the other pr will be made from ehzem's account.


davidharrishmc:
I don't think lui can hit the RV64_rd_corners_lui corners.  Suggest redefining these to be reachable.

PR #111, opened 2024-09-28
Fixed cp_asm_count for compressed instructions

Zain2050:
```
cp_asm_count : coverpoint ins.ins_str == "c.li"  iff (ins.trap == 0 )  {
        option.comment = "Number of times instruction is executed";
        bins count[]  = {1};
    }
```

By changing "c.li" to "li", the bin was getting hit. But this modification was leading to errors because testgen.py uses this to recognize an instruction to generate its tests. This was fixed by 

```
cp_asm_count : coverpoint ins.ins_str == "li" || ins.ins_str == "c.li"  iff (ins.trap == 0 )  {
        option.comment = "Number of times instruction is executed";
        bins count[]  = {1};
    }
```

This made the bin hit and testgen.py correctly recognized "c.li" as the instruction.

### DIFF
--- a/bin/covergroupgen.py
+++ b/bin/covergroupgen.py
@@ -84,7 +84,11 @@ def customizeTemplate(covergroupTemplates, name, arch, instr):
         return ""
     instr_nodot = instr.replace(".", "_")
     template = template.replace("INSTRNODOT", instr_nodot)
-    template = template.replace("INSTR", instr)
+    # This cond is added so that compressed instrs without "c." can also be detected by cp_asm_count
+    if (name == "cp_asm_count" and instr.startswith("c.")):
+        template = template.replace("INSTR", f'{instr[2:]}" || ins.ins_str == "{instr}')
+    else:
+        template = template.replace("INSTR", instr)
     template = template.replace("ARCHUPPER", arch.upper())
     template = template.replace("ARCHCASE", arch)
     template = template.replace("ARCH", arch.lower())
--- a/bin/covergroupgen.py
+++ b/bin/covergroupgen.py
@@ -84,9 +84,9 @@ def customizeTemplate(covergroupTemplates, name, arch, instr):
         return ""
     instr_nodot = instr.replace(".", "_")
     template = template.replace("INSTRNODOT", instr_nodot)
-    # This cond is added so that compressed instrs without "c." can also be detected by cp_asm_count
+    # This cond is added to neglect "c." from compressed instructions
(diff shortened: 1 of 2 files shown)


davidharrishmc:
That technique worked, but it's a nasty hack. The coverpoint now involves a comparison that never happens during real Verilog simulation, and is only used to fool testgen.py? Can you find a cleaner way to solve this by modifying testgen.py rather than the coverpoint?

Zain2050:
Doesn't this OR comparison work in real verilog simulation? Because when I ran **make sim**, cp_asm_count was being hit no matter what the order was
```
cp_asm_count : coverpoint ins.ins_str == "c.li" || ins.ins_str == "li"
cp_asm_count : coverpoint ins.ins_str == "li" || ins.ins_str == "c.li"
```
which made me think that the OR operator was working. 
I just went with the latter one because while reading the coverpoint, regex in testgen.py was reading the latter double quotes for writing its tests.

davidharrishmc:
Your code works, but ins_str is never c.li in the real environment, so it is testing for something that never happens.  A year from now, nobody will remember why you made this hack, and nobody will understand why you are checking both names.  We hope these covergroups will eventually be used by everyone who wants to do RISC-V functional coverage, and they will wonder why the || is needed (and it is not needed for the covergroup, only as a hack to avoid fixing the root problem in testgen).  

One easy fix would be to add an associative array in testgen listing the official compressed instruction name (c.addi, c.addi4spn, c.addi16sp, etc.) and the corresponding name seen during simulation (addi, for all three of these).  Modify testgen to print the appropriate name when printing the cp_asm_count coverpoint.  I think this should be a tiny code change aside from typing out the array.  Another option would be to add a column to the Zc* testplans listing the alternate name. Maybe you have a more graceful idea.

Zain2050:
Now covergroupgen.py is generating 
```
cp_asm_count : coverpoint ins.ins_str == "li"  iff (ins.trap == 0 )  {
        option.comment = "Number of times instruction is executed";
        bins count[]  = {1};
    }
```
I have modified testgen.py in such a way that when it reads an instruction from within a Zc* directory, it will add **"c."** to it and then generate tests accordingly. For example, it will read **'li'** from the coverpoint and generate tests for **'c.li'**.
I hope this approach is better.

davidharrishmc:
I accepted this PR because it works for most cases.  However, it doesn't hit c.addi4spn or c.addi16sp, both of which show up as addi via RVVI.  It would be great if you can think about how to generalize or special case to hit those, and any other compressed instructions that don't just drop the c.

PR #112, opened 2024-09-28
Lui 64 is at 100 percent coverage same as 32 instruction

Abdullah-Mujtaba:
me sher and ehzem we worked together to make this instruction to get 100 percent coverage


davidharrishmc:
I think you forgot to do a git pull upstream main before pushing.  Look at the Files changed tab in your PR.  For example, signedImm6 and c.li are now missing in testgen.py.  Somebody pushed it recently, and I don’t think you were intending to delete it.  Pull the latest repo, fix any merge conflicts, and commit and push your changes again, and your PR will automatically be updated.  I think you have some bugs as well.  In cp_rd_corners_lui, you do a make_rd_corners_lui with xlen=64 and again with xlen=32.  xlen already has a value here, and you only should make the corners for the XLEN being tested.  I don’t understand your make_rd_corners_lui code.  the immediate_values are the same for both xlen except the random one, and the random one doesn’t look random for xlen=64.  Why do they need to differ between rv32 and rv64?  I think you could have left your corners_20bits without commenting it out, and used that for both, if you fix the template for rv64 random lui.  The code would be about a quarter the length of what you have right now.

davidharrishmc:
@Ehzem are you driving this one?  Can you finish it out?

PR #113, opened 2024-09-28
c.slli 100% Coverage

Mysterio-Abdullah:
I looked into the testplan for compressed instructions and added the possible coverpoints. 
cp_rd_sign
cp_rdp  
cp_rdp_toggle_slli      (Exclude bit_1_1  = {32'b???????????????????????????????1}; because its not possible as imm = 0 isn't allowed)
cp_rdp_corners_slli    (Possible rd corners which don't have 1 in Least Significant Bit as they will be replaced by 0 after shifting)
 
 rd is replaced by rdp to emphasize on compressed case.
 
 If any coverpoint is seems missing or if anything can be improved, please let me know.


davidharrishmc:
I think the fourth entry in c_slli_32_corners is only 31 bits, which is confusing to read.  Should it have an extra leading 0?

Mysterio-Abdullah:
Yes Sir, thanks for pointing. Now, I made 32 bit enteries of c_slli_32_corners. 100% coverage has been achieved after Zain's effort.
Sir, please let me know if any other flaw in the approach. If not, can I do such things for all other compressed shifts?

davidharrishmc:
Nice!  Looks good.

Stepping back, this would be a good time to think about whether all the features of slli are tested.  Looking at the covergroup, the obvious feature missing is that we don't exercise the shift amount.  I think we need a coverpoint like RV{32/64}_cp_imm_shift.  Shift amount is 5/6 bits according to XLEN.

Think about whether there's anything else we should be exercising.  When you are satisfied, sounds good to extend to the other shifts!

PR #115, opened 2024-09-29
improved cp_asm_count for compressed instructions

Zain2050:
@davidharrishmc Thank you for pointing out that cp_asm_count for c.addi16sp and c.addi4spn are not being hit. 
I have added a dictionary in covergroupgen.py which handles all special cases like c.addi16sp, etc. We can add all special instructions in that dictionary and strip **'c.'** for other compressed instructions.
As for testgen.py, I have changed my approach to grab the compressed instruction's real name. 
```
    option.comment = "c.addi4spn";
    cp_asm_count : coverpoint ins.ins_str == "addi"  iff (ins.trap == 0 )  {
        option.comment = "Number of times instruction is executed";
        bins count[]  = {1};
    }
```
**curinstr** was being fetched from the cp_asm_count line. I have adjusted the code such that if we are in a Zc* directory, we will fetch the instruction name from the previous line (from option.comment). This is a more general approach as compared to adding a **'c.'** before every instruction.

If you have any suggestion, let me know.

### DIFF
--- a/bin/covergroupgen.py
+++ b/bin/covergroupgen.py
@@ -84,9 +84,14 @@ def customizeTemplate(covergroupTemplates, name, arch, instr):
         return ""
     instr_nodot = instr.replace(".", "_")
     template = template.replace("INSTRNODOT", instr_nodot)
-    # This cond is added to neglect "c." from compressed instructions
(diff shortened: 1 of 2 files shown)


davidharrishmc:
This is graceful.  Nice work!

PR #118, opened 2024-09-29
Zca shifts 100%

Mysterio-Abdullah:
Only shift_imm is left for all. It's not working even for base instructions, but I'll try to look deeply into it.


davidharrishmc:
At line 97, it looks like the immval is constrained from 0-30.  Don’t you want 1-31, or more generally 1-(XLEN-1)?

Why do we need cp_rdp_toggle?  Could we use cp_rd_toggle equivalently?  Same with cp_rdp_sign

davidharrishmc:
Why are the values strings in lines 96-97?
Still open question about the rdp coverpoints.

Mysterio-Abdullah:
Sorry for the late rsponse. I slept after making PR.

shiftImm function will return the range (0-31):
imm = imm % xlen

And I made an if statement to exclude immediate value 0. So, I beleive it's fine. But it can be generalized by giving the arguments shiftImm(immval, xlen).

Me and Sir Umer were looking into the test plan and inspired by cp_rdp instead of cp_rd for compressed register set. So, we decided to keep things separate for compressed register set and base register set by making their separate templates.
For example:
cp_rd_toggle, cp_rd_corners, cp_rd_sign for base instructions and cp_rdp_toggle, cp_rdp_sign for compressed instructions
But, If you have any suggestions, please let me know.

davidharrishmc:
I was wondering why the shiftImm comparisons are to strings rather than integers.  Can you drop the quotes?

Are the rdp coverpoints identical to the rd coverpoints?  I know they are used for 8 vs. 32 registers, but is the code identical?  If two coverpoints have identical code, I'd prefer they have an identical name to reduce the learning curve for the reader.

Mysterio-Abdullah:
The comparisons are to strings because the function shiftImm returning the str(immediate). That's why we can't drop the quotes.

The the templates & coverpoints for rd & rdp are almost identical for most of the time. But separating both will give you some sort of ease to modify the testgen.py to generate the test for each case. But it's possible that we can use same. Let me know, if I should?

PR #127, opened 2024-10-02
c.mv 100% coverage.

Mysterio-Abdullah:


davidharrishmc:
What happens if we use cp_rd instead of cp_rd_nx0?  Is it possible to generate instructions that write x0?  That's a nice case to verify if it is possible.  If not possible, what shows up in objdump for the instructions that attempted to write x0?

Mysterio-Abdullah:
Yes Sir, it is posiible to generate instruction that write x0. It does't show any error of illegal operands and the bin get hits. I excluded x0, just because of the restriction.

li x0, 0x735bdc92 # initialize rd to a random value that should get changed; helps covering rd_toggle
800010e6:	735be037          	lui	zero,0x735be
800010ea:	c9200013          	li	zero,-878
li x11, 0x2fdae112
800010ee:	2fdae5b7          	lui	a1,0x2fdae
800010f2:	11258593          	addi	a1,a1,274 # 2fdae112 <NUMTESTS+0x2fdae111>
c.mv x0, x11 # perform operation
800010f6:	802e                	c.mv	zero,a1

davidharrishmc:
This came up in the 10x channel.  If you can include the x0 bin for rd, let's do so.  That potentially catches certain implementation bugs.

Mysterio-Abdullah:
Already done Sir, in last commit.

davidharrishmc:
I was just looking in the wrong column.  Great!

PR #129, opened 2024-10-03
100% coverage on fadd.s and fsub.s

AnonymousVikram:


davidharrishmc:
rs2 += 1 is ill-defined if rs2 = 31.  
Make_cp_csr_frm replicates a block of code from writeCovVector.  It is better to extend writeCovVector than to paste the code here.  For example, you have the same bug li x4; sw x3 in the pasted code that Ahylissa fixed last night.
I’m not entirely clear on the operation of the rounding mode tests.  It looks like the code sometimes steps through the rounding modes as an instruction suffix, and sometimes with the CSR.  When you sweep rounding mode, I think you want to hit both the static and dynamic rounding modes all at once.  In particular, a cross-product involving FRM should hit rounding modes 0-4 and 7, with all the 5 different fcsr.frm values when mode is 7.

davidharrishmc:
The previous version set fcsr.frm back to RNE after tests.  That seems to be missing now.  I think it's worth restoring, with one fcsr write after all of the mode tests.  Or put RNE mode last so it automatically ends up back in the default mode after the frm corners.

PR #136, opened 2024-10-03
improved coverage of c.sub

hamza-1821:
improved coverage to 98%
files changed :
1) testplan
2) testgen.py
3) sample_RV32Zca.txt


davidharrishmc:
In the cadd/cmv case, is it necessary to avoid rd = 0?  If we can test that case, it would be beneficial for robust coverage.

Your catype code is out of place.  The initialize rd to random ought to happen before any of the different test types.  You put it in an if/elf, which requires that you duplicate it, and it makes your code asymmetric with all the other types.  I’d suggest moving your code down to near the other c*types in the elif block.

The catype code in make_rd_corners is not robust.  At best it will work for c.sub, not c.or, c.xor, c.a.  The v*2 doesn’t work for c.sub if v has a 1 in the most significant bit.  I don’t think cp_rd_corners is a terribly useful coverpoint for Zca, or worth the trouble of thinking how to hit a particular rd.  How about just hitting corners for the two inputs?

If you add the other ca-type instructions to the catype = line and fix it up in RV32Zca, you’ll test all four rather than just one.

hamza-1821:
Right ! 
I think I might be a little lost on cadd/cmv here. I’ve made the changes to the code as you suggested regarding the random initialization and ca-type instructions. However, I’m not currently working on the c.add/c.mv case, so I wanted to clarify if you were referring to those instructions or if this comment was related to a different area of the project?

davidharrishmc:
Your PR touched lines 104-5 related to c.add and c.mv.  If you arent' intending to change those lines, you can just revert them.  

image

PR #138, opened 2024-10-03
93.26% jalr covered

Daniyal-R-A:


davidharrishmc:
Could you add cp_rs1 for jal?  And think about whether any other coverpoints are necessary to fully exercise the instruction?

PR #143, opened 2024-10-03
Problem was related to pseudoinstructions

Shahjahan07:
So made the relevant change in covergroupgen.py


davidharrishmc:
For Snez

davidharrishmc:
Please take a look at the improved covergroupgen pseudoinstruction code.  There is now code able to handle add_rs1_0, which snez will need to use (see sub as an example).  You will need to adjust your code.

You will probably want to implement the sgtz pseudoinstruction at the same time.  It is almost the same:

sgtz rd, rs = slt rd, x0, rs        Set if > zero

Shahjahan07:
Sir I need to look into this, and I'd get back to you.

PR #145, opened 2024-10-03
c.lw 100% Coverage.

Mysterio-Abdullah:
Managed to get 100% coverage of "c.lw"
Luckily, 100% coverage of c.sw has been achieved from the test generated for "c.lw" because I was using c.sw in them.
Should I generate separate tests for "c.sw" ?

### DIFF
--- a/bin/testgen.py
+++ b/bin/testgen.py
@@ -321,11 +321,11 @@ def make_rs2(test, xlen):
     writeCovVector(desc, rs1, r, rd, rs1val, rs2val, immval, rdval, test, xlen)
 
 def make_rd_rs1(test, xlen):
-    rng = range(8, 16) if test in cltype else range(32)
-    for r in rng:
-        [rs1, rs2, rd, rs1val, rs2val, immval, rdval] = randomize()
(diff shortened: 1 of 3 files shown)


davidharrishmc:
Terrific!

Please do generate separate tests for c.sw.  We would like to be able to test each instruction independently even though some of them naturally do cover others.

davidharrishmc:
This looks good.  There's a simplification in make_rd_rs1.  The for loops appear to be cut and paste of each other, only differing in the range.  Could you set a variable rng to either range(8,16) or range(32) based on (test in cltype), and then only have one copy of the loop, based on for r in rng?

Mysterio-Abdullah:
Ok Sure. I'll generate individual tests for "c.sw".
Also, I simplified the make_rd_rs1. Thanks for pointing.

davidharrishmc:
Minor issue -- the make_cmp_rd_rs1 function now has the body double-indented.

Mysterio-Abdullah:
Sorry for that. I think everything is fine now.

PR #148, opened 2024-10-04
c.sw 100% coverage

Mysterio-Abdullah:
100% coverage of c.sw achieved after unchecking all the coverpoints of c.lw to ensure that its each bin is getting hit because of its own tests. i.e,

Covergroup instance obj_c_lw                             0.00%        100          -    Uncovered
Covergroup instance obj_c_sw                         100.00%        100          -    Covered


Mysterio-Abdullah:
Please let me know if anything can be improved :)

davidharrishmc:
Looks reasonable.  Why have the if check, instead of just legalizing the register numbers?

Mysterio-Abdullah:
Sir, I'm already legalizing the rs1 and rs2 in make_rs2p. If i try to legalize it again by using legalizecomp function, it will not provide all rs2p from 8 to 15.

Mysterio-Abdullah:
I also removed cp_rs1_toggle for both c.sw and c.lw because it seems useless as we are using rs1 as a base address to load and store.

davidharrishmc:
I presume you are saying this from experience, but I’m not seeing the problem on casual inspection.  If line 157 were gone, why wouldn’t we still cover all rs2p from 8 to 15?

Mysterio-Abdullah:
Apologise, I also went through it again, and seems everything fine without it.

PR #151, opened 2024-10-05
c.or & c.andi at 100%

Zain2050:
I made a new coverpoint cp_imm6_corners in RV32Zca.csv, which checks 6 bit imm corners without cross coverage checking with any other value. 

I am checking cp_rd_corners for c.or, but leaving it in case of c.andi. I cannot think any way to make it hit using a 6 bit immediate. 
Coverpoints for c.andi are:
cp_asm_count, cp_rd_sign, cp_imm_sign, cp_imm6_corners, cp_rdp, cp_rd_toggle

Is it okay to leave cp_rd_corners out or should I try to hit it? And if I should be hitting any other coverpoints for c.andi, please let me know.


davidharrishmc:
Your plan to skip cp_rd_corners for c.andi is sensible.  You have other ways to exercise all the possibilities.

There are a number of instructions with the same formats.  Instead of just doing c.or and c.andi, can you aim to do all the other ones of the same formats at the same time?  And can you look at them for both RV32 and RV64 at the same time?

I’d really like you to think about what coverpoints are needed to prove the instruction works, rather than for me to tell you. If you have questions about a coverpoint, I’m happy to discuss.

Instead of needing separate make_cp_imm6_corners, cp_imm12_corners, etc., can you have a single make_cp_imm_corners function and pass it different lists of corners (e.g. corners_imm_6bits, etc.)  Code is cleaner if you avoid duplication.

There are some asymmetries in the cp_imm_6 corners.  Can you make the choices more symmetric?

Adding a new cp_imm6_corners column to a testplan is painful.  Instead of doing that, could you reuse the existing cr_rs1_imm_corners, which would be better anyway because it will give you the cross-product?  Instead of putting a x in the column can you put a 6bit there? Just like c.li?  Then you could do similar if you ever need immediates of different sizes.

Zain2050:
I added imm_corners as a function argument to **make_cr_rs1_imm_corners** and **make_cp_imm_corners**.
Removed the newly added column and instead added **6bit** to **cr_rs1_imm_corners**, checking cross coverage for it.
Made **cp_imm6_corners** symmetrical.

About the other instructions of the same format, I think you probably mean instructions like (c.and, c.xor, etc). I can work on them, but they are assigned to Hamza and Ahmad.  I was thinking of working on c.bnez, c.beqz, and then moving onto RV64Zca. However, if you want me to work on c.and, c.xor, etc., I can absolutely do it.

PR #152, opened 2024-10-05
100% Coverage of LUI for rv64i

Ehzem:
Fixed the bins of cp_rd_corners_lui6_4, 
Stopped testing cp_rd_toggle cover point.


davidharrishmc:
Please keep the indentations consistent.  Make_rd_corners_lui has larger indentation than all the other functions.

I can’t see what else has changed in make_rd_corners_lui.  Please don’t commit code changes unless you actually need them.

As posted in the 10x channel, remember that when you need to have a separate template for RV64 vs. RV32, put on a prefix to distinguish them.  For example, if cp_rd_corners_lui.txt needs to be diffent because the expected corners are 64 vs. 32 bits, rename the old 32-bit cp_rd_corners_lui.txt to RV32_cp_rd_corners_lui.txt and create a new RV64_cp_rd_corners_lui.txt.  Both files should contain a coverpoint named cp_rd_corners_lui; don’t append a 64 suffix on one of them, for example.  

I don’t understand why there is a corners_40bits.  lui takes a 20-bit argument for both RV32 and RV64.  There is nothing interesting in bits 39:20 – it’s just sign extension.

PR #153, opened 2024-10-05
c.slli, c.srli, c.srai, c.lw, c.ld, c.sw, c.sd and c.mv 100% Coverage for RV64Zca.

Mysterio-Abdullah:
There's some issue maybe in doing regression-wally --fcov. I got 100% coverage of c.slli for RV64Zca but only when I do make sim.
If you have any clue, what the root cause is, it will help me to resolve it too. But, atleast I made the required changes to get 100% coverage of c.slli


Mysterio-Abdullah:
Now, everything is fine. Please let me know, if anything can be improved.

davidharrishmc:
For cstype, I think mul should be 4 for sw and 8 for sd, not 4 for xlen=32 and 8 for xlen=64.  Also, the immval is the thing that should be unsignedImm6, not immval*mul.  Otherwise you aren’t testing the full range of immediates.  It’s interesting that you are getting full coverage.  There must be some bug in our coverage file that isn’t detecting that you aren’t exercising the full range of immediates.  Can you look into it?

Everything else looks good.  Way to go for so much coverage!

Mysterio-Abdullah:
Yes, you're right sir. It should be (ZeroExt(imm)*mul)(rs1) not (ZeroExt(imm*mul))(rs1). Also, it's a good catch that mul doesn't depend on xlen, but instructions.

PR #156, opened 2024-10-05
Coverage for SRAIW 64, 100 percent

Abdullah-Mujtaba:
Made new corner cases for sraiw, that work correctly. Immediate was being checked for the upper 32 bins of this instruction as well which is not valid as sraiw only allows valid shifts upto 32 bits.


davidharrishmc:
Why do you need different immediate values for xlen=64 and xlen = 32?  Can you pick the 4AE20 random value (01000 doesn’t really look random) and elimiante the xlen=64 code?  The immediate values are usually defined later in testgen.py where other corners are defined.  Could you do the same.  It’s ok not to have the fancy descriptions.  Let’s keep the code simple and succinct.  Looking further, it seems the old make_rd_corners_lui might be sufficient.  Is there a reason you were changing it?

There's now a merge conflict as well after another PR was merged.

PR #161, opened 2024-10-06
Re-add frm and fcsr.frm tests

AnonymousVikram:
Note: fmul and fdiv were also added, but these need to be slightly modified after I complete the refactor of fflags


davidharrishmc:
Can you change the order of the entries in csrFrm to end with rne so you don't have to reset frm externally?

AnonymousVikram:
> Can you change the order of the entries in csrFrm to end with rne so you don't have to reset frm externally?

Hadn't thought of that. Thank you! I've implemented it now.

davidharrishmc:
Something is funky.  You have 20 files in the PR, with most of the commits coming from somebody else?  Can you redo with just your changes?

PR #163, opened 2024-10-06
Added cp_imm_mul coverpoint for Zca

Mysterio-Abdullah:
I also put "x" in the column of cp_imm_mul for different instructions for which we have to test multiple of immediates. It will help someone to emphasize that we have to check this coverpoint also. Then he will modify it according to its requirement e.g, cp_imm_mul_8 and cp_imm_mul_16.


davidharrishmc:
I suggest refactoring the cp_imm_mul coverpoints.  They don’t depend on RV32 vs. 64, but rather on the multiplier.  Looks like you need 4, 8, and sign-extended 16.   I think they could be called cp_imm_mul_4, cp_imm_mul_8, and cp_imm_mul_16.  The comment about Immediate multiples of 8 for c.sw isn’t accurate – multiples of 8 are for sd.  They also don’t go up to 504.  Simplify the comment and make sure it is always true.

The coverpoint could always be called cp_imm_mul, and you can rely on the instruction rather than the name of the coverpoint when you figure out how to produce the immediate in testgen.

Mysterio-Abdullah:
Thanks for pointing some mistakes. Now, I made only one generic template cp_imm_mul. And also remove that silly comment. Please have a look now and let me know if still something missing.

PR #164, opened 2024-10-06
100% coverage for c.and, 99% coverage for c.xor

hamza-1821:
Incase of c.xor, I am not hitting a specific bin in which rs2 = rd. There can be an if/else condition made to avoid both registers being same, but it will not be robust.


davidharrishmc:
You're right rs2 cannot be rd for cp_rd_corners because the result will be 0.  You'll need to exclude that choice of rd in cp_rd_corners.  There are lots of other instructions with a while (rd == rs2) check to pick an alternative register; just do the same.

PR #172, opened 2024-10-07
100% coverage for RV32 & 64 c.addi16sp

ahmad101-sudo:
modified:   bin/testgen.py
        modified:   templates/sample_RV32Zca.txt
        modified:   templates/sample_RV64Zca.txt
        modified:   testplans/RV32Zca.csv
        modified:   testplans/RV64Zca.csv
        new file:   templates/RV32_cp_imm_ones_zeros_addi16sp.txt
        new file:   templates/RV64_cp_imm_ones_zeros_addi16sp.txt
        new file:   templates/cp_imm_mul_addi16sp.txt


davidharrishmc:
The addi16sp immediate is 6 bits.  Why are there 9 1s in the cp_imm_ones_zeros coverpoint?

What is cp_imm_ones_zeros covering that cp_imm_mul_addi16sp doesn’t already cover?

ahmad101-sudo:
The bins "wildcard bins bit_10_1" to "wildcard bins bit_12_1" in  cp_imm_ones_zeros coverpoint are hitting as addi16sp has sign extended 6 bit immediate. The cp_imm_mul_addi16sp is generating the test for every case and the cp_imm_ones_zeros coverpoint is telling us about the number of times each case is getting hit.

davidharrishmc:
In my opinion, cp_imm_ones_zeros won't detect any bugs that cp_imm_mul_addi16sp doesn't catch, and could be dropped by removing the x.  If you feel strongly, you can leave it in.  Those wildcard bins also aren't catching any bugs and could be dropped.  In any event, we now have a merge conflict you'll need to resolve.

PR #173, opened 2024-10-07
Sraiw RV64 at 100 percent, Srliw at 94.27 percent

Abdullah-Mujtaba:
Made new corner cases for Sraiw and added a new test for immediate for the word instructions, which led to Srliw's coverage to reach 94.27.


davidharrishmc:
Why is make_imm_shift_w needed?  It looks very similar to make_imm_shift, aside from a typo in the description. :)

You don’t need a cp_imm_shift_w column in the .csv.  Instead, put a w instead of x in the cp_imm_shift column.

Abdullah-Mujtaba:
> Why is make_imm_shift_w needed? It looks very similar to make_imm_shift, aside from a typo in the description. :)
> 
> You don’t need a cp_imm_shift_w column in the .csv. Instead, put a w instead of x in the cp_imm_shift column.

PR #176, opened 2024-10-08
Created half precision fp corner values and added support for Zfh extension

coreyqh:
* adds half precision fp corner values
* adds double and half precision cross products 
* updates the RV{32,64}Zfh testplans
* modifies testgen to integrate the Zfh extension without breaking sim and having to remove tests
* regression-wally --fcov runs without failure
* I didn't find any lost coverage


davidharrishmc:
The Zfh SNaN bins include infinity.  They need to start 1 higher to exclude infinity.  Please double check if this applies to any other precisions.

coreyqh:
Thanks for catching that! I noticed the same issue is present in the single precision coverpoints and the double precision coverpoints had a completely wrong value with a 7 prepended. I fixed each of these coverpoints as well as testgen.

PR #180, opened 2024-10-09
c.sb & c.sh 100% Coverage

Mysterio-Abdullah:
Exclude checks from coverpoints like cp_rs1_corners, cp_rs1_toggle as rs1 is just a base address "scratch". Also exclude cp_fdp and cp_fs2p. Also made a change in makefile which is required. Please let me know, if anything can be improved.


davidharrishmc:
@Mysterio-Abdullah I reverted your PR #180 while looking for a bug that broke regression.  It turned out to be in a different PR.  Can you reissue this PR?

PR #194, opened 2024-10-12
Near full coverage of fmv.x.w using alias fmv.x.s 

coreyqh:
Hitting 98.87% of bins. 
Only missing asm_count bin, which may not be possible to hit without an imperas fix.
This can be easily extended to other fmv* instructions should they also be aliased to depreciated names.


davidharrishmc:
How about drop asm_count, or have a modified version of it that recognizes the old name?

coreyqh:
I'll get to work on that, full coverage very soon!

PR #195, opened 2024-10-13
completed c.subw

hamza-1821:
Added a new template for cp_rd_corners for c.subw. Some of the bins needed sign extension.


Mysterio-Abdullah:
Can I add something? Instead of making another template cp_rd_corners_subw.txt, you can use exisiting template RV64_cp_rd_corners_32bit.txt and instead of passing corners you added, you can pass existing corners "corners_32bits" in testgen.py.

Let me know if it works :)

hamza-1821:
Good thought ! might've missed them.

Mysterio-Abdullah:
Now, its better to delete the template you wrote "cp_rd_corners_subw.txt" ig.

davidharrishmc:
cp_rd_toggle_lui sounds like it should apply to all lui instructions, but I think it only applies to c.lui.  Can you change the name accordingly?

@Mysterio-Abdullah thank you for your review on this too!

hamza-1821:
Done.

hamza-1821:
For c.lui both in 32 and 64, there is one bin in cp_rd (bin auto [x2]) that remains, because x2 is invalid for c.lui so it cannot be hit. Do i need to create a template for c.lui for this 1 bin too ?

davidharrishmc:
Looks like it.

davidharrishmc:
Shouldn't the clui templates have 6 ones instead of just 5?  The immediate is 6 bits.

hamza-1821:
Yes ! but we are providing c.lui with 5 bit Uimmediate value instead of 6 (it generates error for 6 bit for some reason).

davidharrishmc:
I posted about this in the 10x channel.  It's really important to find a way to test all of the bits, rather than skip one.  Never write a coverpoint that hides something that actually needs testing. Regular lui may have a similar challenge.  Do we need a signed rather than unsigned immediate?

PR #197, opened 2024-10-13
RV64/32Zcb c.zext.b Done

Zain2050:
Made some new templates for 8-bit zero extended registers.


davidharrishmc:
Can you use the cp_rd_corners_lbu template here instead if it is identical?

PR #198, opened 2024-10-13
Resolved some issues from PR#197

Mysterio-Abdullah:
PR # 197 removed a coverpoint "cr_rs1_imm" from .csv, but unfortunately, he forgot to delete the commas of its column that affected the coverage of all other Zcb instructions. Moreover, it is getting the coverage of cp_rs1, cp_rs1_sign and cp_rs1_toggle while it's not required as c.zext.b has no rs1 field.


Zain2050:
cr_rs1_imm was written twice, therefore I removed one and possibly forgot to remove the commas. 
Why have you removed rs1 from c.zext.b and removing all the coverpoints of rs1? rd is also rs1. rs1 is being used to check rd values before performing the operation. Don't change these, only fix the commas. Thank You!

PR #201, opened 2024-10-14
Getting missing coverage of auipc, c.add, c.and, c.xor, c.sub

Mysterio-Abdullah:


davidharrishmc:
Can you make the title a bit more specific about which instructions are covered?  Is it full or partial coverage?

Mysterio-Abdullah:
Compressed instructions were at 95+%, only two three bins were not hitting for each instruction because of rd = rs2 in rd_corners.

Mysterio-Abdullah:
Also, cp_rd_corners_auipc was actually cp_rd_toggle_auipc. And remove the column of cp_rd_corners_auipc from .csv file because it's not a standard now.

PR #205, opened 2024-10-14
RV64ZcbZba Done

Zain2050:
While doing this, I faced a problem. **ZcbZba** is only for RV64, and testgen.py was trying to find RV32 coverage.svh file for it. As it couldn't find it, testgen.py was throwing an error and stopping execution. 
I made a modification in testgen.py, which we can use for extensions that are only made for RV64 or RV32.


davidharrishmc:
nice solution!

PR #209, opened 2024-10-15
100% Coverage of RV32I jalr instruction  

Daniyal-R-A:


davidharrishmc:
Please remove the Makefile from the PR.  Let's not check in Makefile changes unless everyone else needs them too.

PR #217, opened 2024-10-16
RV32 c.j & RV32/64 c.lwsp & RV64 c.ldsp

ahmad101-sudo:


davidharrishmc:
What is the purpose of the beq t1, t2, label2 code?  It seems it would never be executed.
What is the purpose of the nop statements after the labels?

if (test == "c.j"):
lines = lines + "li t2,0\n"
lines = lines + "li t1,1\n"
lines = lines + "label2:\n"
lines = lines + " nop\n"
lines = lines + "beq t1, t2, label3\n" 
lines = lines + "beq t1, t2, label2\n" 
lines = lines + test + " label1\n"
lines = lines + "label1:\n"
lines = lines + " nop\n"
lines = lines + "li t1,0\n"
lines = lines + test + " label2\n"
lines = lines + "label3:\n"


In the following code, why 32?  Isn’t the offset a 6-bit immediate?
if (test == "c.lwsp"):
rng = range(32)

ahmad101-sudo:
Currently we are using 5bit unsigned immediate as we are getting errors of illegal operands for 6bit unsigned immediate.
Hamza is looking into that issue hopefully it will be resolved.
You are right  beq t1, t2, label2  is pointless. 
i will remove it in next commit.

davidharrishmc:
What is the function of the nop s?

It's dangerous to be reporting 100% coverage when one of the bits of the immediate is not being exercised.  Please at least change the coverpoint so we don't get a deceptive report.

PR #218, opened 2024-10-16
c.j , c.lwsp , c.ldsp

ahmad101-sudo:
Sir, now I reviewed all the instructions that use 6 bit immediate unsigned value. Basically, when we started getting coverage of  Zca, there's a function named as UnsignedImm6, but it was providing the 5 bit unsigned immediate value. And also you added a comment there that "This is causing assembler error right now for instructions." So, for all the instructions, we were using the same function you wrote which was supposed to give 6 bit unsigned Imm value. But now I made a new function "ZextImm6" for all the instructions that use 6bit excluding c.lui which actually causes assemble error. You can review it now and let me know if anything can be improved.


ahmad101-sudo:
Yes Sir, it fixed all the coverpoints templates of the instructions that use 6 bit unsigned value. Now, only c.lui is giving assemble error. All the other instructions, which are supposed to get 6 bit unsigned immediate are now actually getting 6 bits.

PR #222, opened 2024-10-16
100% coverage for FR4-type instructions

HamzaYusufJamal:
@coreyqh some changes needed to be made to the RV32F and RV64F csvs to get to 100%, please propagate these to the other related csv's


coreyqh:
It looks like this PR deleted some cross product coverpoints for other floating-point precisions. Was this intentional? I am pretty sure we will need these for the other extensions.

PR #231, opened 2024-10-17
100% coverage of addw for RV64I

TheUnparalleledDragon:
Use different template for addw rd_corners

RV64_cp_rd_corners_32bit


davidharrishmc:
Can you do this for all of the RV64w tests that are missing rd_corners coverage, instead of just addw?

PR #236, opened 2024-10-17
100% coverage of jal64I, jalr64I and subw64I

Daniyal-R-A:


ahmadusman86:
Dear Daniyal 

I am unable to see jalr (jalr 64I) and subw command in the text.

jal is UJ format - This command requires checking of bits from 12 to 31, whereas I see changes from 2 to 12. Kindly explain this. 
jalr is I format  - This command requires checking of bits from 0 to 11. This is missing in the text.
subw command - There is no information on the changes available in the text. 

Please explain.

Daniyal-R-A:
Dear Professor,
The issue was also occurring while testing jal & jalr 32I.
The Logical purpose of removing bin 0_1 and 1_1 from cp_rd_toggle is because it checks the (ins.current.rd_val). We know that a single instructions takes 4 bytes. Therefore our PC must increment at least by 4 or some multiple of 4. bins 0_1 and 1_1 were failing because it is not the multiple of 4 therefore they must be removed in order to achieve full coverage.
The purpose of testing for fewer bins and not all (64+64) bins is because of memory constraints of simulation. We can assume that other bins can be covered by other instructions which are generating the value randomly.

The changes for subw is done in covergen.py file because it was generating a pseudo instruction of "negw", when rs1 = x0, for which no sample was being generated in RV64I_Coverage.svh. Hence by adding the lines 
 " if name.startswith('sample_') and instr == 'subw':
          template += template.replace(instr, 'negw',1).replace("add_rs1","add_rs1_0",1).replace("add_rs2(2)", "add_rs2(1)")        " 
in coverage file we can cover the whole instruction.

Daniyal-R-A:
Actually, we havent been instructed to make a separate PR for each instruction. However, I have made separate commits for Jump type and subw instructions.

davidharrishmc:
Jumps allow the target address to be a multiple of 2 rather than 4 because compressed instructions can be two bytes.  However, in RV{32/64}I, there are no compressed instructions and no way to hit addresses that are not multiples of 4.

I think the best solution is to add uncompressed jal and jalr rows to the Zca tests.  These should have two coverpoints consisting of rd[1] = 1 and imm[1] = 1.  jalr also needs to be able to test a case where rs1+imm is odd; the bit [0] should be ignored. We need to test both rs1[0] = 1, imm[0] = 1, and rs1[0] + imm[0] = 1.  See the unprivileged spec about jalr. That way these features can be tested when compressed instructions are supported, and everything else is tested in the usual I tests.

Can you work with the UET team on this?  Perhaps create a proposed PR and review it with them?

UmerShahidengr:
@Zain2050 can you please look into this, I think you have worked on jal instructions. There is a missing coverpoint which can be covered in Zca

davidharrishmc:
Moved this feedback on uncompressed branch/jump to Issue #258.

PR #237, opened 2024-10-17
c.swsp & c.sdsp 100% Coverage.

Mysterio-Abdullah:
I got 100% coverage of RV32/64 Zca c.swsp and RV64 Zca c.sdsp after unchecking all the coverpoints of c.sw, c.lw, c.lwsp, c.ldsp to make sure that we are getting their coverage because of their own tests. The coverage were initially dropped because of the randomness.


UmerShahidengr (approved):
Looks good to go.

PR #238, opened 2024-10-18
Fixed 0% coverage of c.jalr & c.jr

Zain2050:
c.jalr & c.jr were giving 0% coverage despite writing correct tests. Hamza was initially working on them and thought that there was some issue with the jump addresses. 
I started working on it and found out that c.jalr was printing an error.

`SystemVerilog Functional Coverage: get_gpr_reg() not found gpr`

First I thought that this error was being caused within the **Zca** covergroup, but after thorough searching I came to know that it was being caused in the **RV32I** covergroup.
As c.jalr is passed as jalr, it was getting sampled in RV32I covergroup. However, it was causing an error probably because it was searching for 2 registers (rd & rs1) while c.jalr has only 1 register (rs1).  I added a condition in **sample_JR.txt** so that compressed jalr cannot interfere with it.
Now I'm getting partial coverage. As the issue is fixed, I'll soon get full coverage.


davidharrishmc:
There are some tricky issues here.  I think your solution works for c.jr rs1.  If you find any subtle problems, keep us posted.

There's also an uncompressed form jr that sometimes but not always has an offset.  testgen creates a form with an offset.  But in tests/riscof/work/wally-riscv-arch-test/rv64i_m/privilege/src/WALLY-mmu-sv39-svadu-svnapot-svpbmt-01.S/ref/ref.elf.objdump at 80000870, there is a jr t5 with no offset.  This seems to be throwing off the get_gpr_reg because we don't know whether the register will be in position 0 or 1.  I don't have a solution for this yet.

Zain2050:
Interesting! I'll look into it.

Zain2050:
So, the problem was also occurring with jalr when jalr was used without an offset. I managed to find a solution. 
ins.ops has the following form in case of an offset:

`{'{key:"2", val:"2"}, '{key:"x4", val:""}, '{key:"", val:""}, '{key:"", val:""}, '{key:"", val:""}, '{key:"", val:""}}`

and without an offset:

`{'{key:"x4", val:""}, '{key:"", val:""}, '{key:"", val:""}, '{key:"", val:""}, '{key:"", val:""}, '{key:"", val:""}}`

I have added conditions to check these parameters and choose destination and source registers accordingly.
Both jalr and jr had this problem. This solution is also working for c.jalr & c.jr I made a new sample template for jr instead of deleting and replacing lines of the sample_JR template in covergroupgen.py. 
In RISCV_instruction_base.svh, I added a new function to be used for c.jalr.

davidharrishmc:
Awesome!

davidharrishmc:
Your new add_rd_1 function is somewhat different than add_rd in that it drops rd_val_pre and changes rd_val.  Are you certain this is correct?

 virtual function void add_rd(int offset);
        current.has_rd = 1;
        current.rd = ops[offset].key;
        current.rd_val = current.x_wdata[get_gpr_num(ops[offset].key)];
        current.rd_val_pre = prev.x_wdata[get_gpr_num(ops[offset].key)];
    endfunction

    virtual function void add_rd_1();
        current.has_rd = 1;
        current.rd = "x1";
        current.rd_val = prev.x_wdata[get_gpr_num("x1")];
    endfunction

UmerShahidengr (approved):
Good to go

UmerShahidengr:
@Zain2050 the PR has merge conflict, please remove the conflict, and I will add another review on your latest commit.

Zain2050:
The conflict was caused by the latest PR in which pseudo-instructions were commented out. As psuedo-instructions are not needed, therefore I removed my changes in covergroupgen.py and kept it same as upstream.

PR #248, opened 2024-10-24
D and ZfhD simulating

coreyqh:
* Major 64F bug fix (No longer occasionally hangs regression)
* D and ZfhD simulating will majority coverage (PR coming in cvw to get them simulating for everyone) 
* Minor testplan tweaks, referencing the RV spec


davidharrishmc:
The signedImm12 code and its application is a mess and needs rethinking.

At line 35, It’s cleaner to code “if storeOffset: A else B” rather than “if not storeOffset: B else A”  

Line 39 checks if (storeOffset), which will never happen because line 35 already said no StoreOffset.

StoreOffset is a misleading name because it should also apply to loads and jalr.

Where there is one rat, there are many rats. Anywhere signedImm12 is used in an offset, you need to use your new code.  This includes lines 245/250/251, 318/319, 365/366 and maybe 719-723.

coreyqh:
Refactored based on these comments and applied these changes to other instances of immediate offsets.

PR #255, opened 2024-10-25
Fixed missing coverage of c.li,c.lwsp,c.ldsp 

Zain2050:
c.li, c.lwsp and c.ldsp were at 100%, but Hamza's latest PR #247 had a minor problem that dropped coverage of these instructions.  I found the problem and made changes appropriately. Now they're back at 100.


UmerShahidengr (approved):
Good to go

PR #266, opened 2024-10-30
Zicsr.csv for RV32 and RV64

ahlyssa:


davidharrishmc:
@ahlyssa you have changed a whole bunch of files.  I don't think you intended to change most of them.  Always look at the Files Changed tab when you make your PR to check what you are trying to put into the repo, and only change the ones you really need.  I think it will be easiest if you take a note of what you intend to be changing, then cancel this PR and checkout a clean version of the repo, and reenter your intended changes.

PR #269, opened 2024-10-31
cp_fclass

ahlyssa:
- template to include cp_fclass for fclass


davidharrishmc:
This needs tests to go with it to hit the cover point.

davidharrishmc:
fclass should set exactly one bit in rd, per the unprivileged spec.  The coverpoint shouldn't have ?; they should have 9 0's and one 1 walking through each of the 10 possibilities.

PR #272, opened 2024-10-31
100% coverage of sh rv32i and sd rv64i instruction

Ehzem:
Issue was in cp_rs2 when rs1 = rs2, rs2 was changed. So added a condition which checks and based on cover point if cp_rs1 is being tested and conflict occurs rs2 is changed, if cp_rs2 is being tested and conflict occurs rs1 is changed.


davidharrishmc:
The sh fix has a few problems.  It can cause rs1 to become 0, which will not work for a base address for memory.  It’s also clumsy code to have to look at desc[0:5].

This issue has already been solved more gracefully in make_fs2 (see lines 552-553).  If you put the same code in make_rs2, you should avoid the issue.

    while (r == rs1):
      rs1 = randint(1,31)

PR #281, opened 2024-11-05
Added cross coverage of fs1 and frm for single source fp instructions

coreyqh:
* Ensured that every corner value is rounded with each rounding mode
* No loss in fround* coverage
* This coverpoint may be applicable to other instructions, pending review from the HMC team


davidharrishmc:
I'd suggest applying this to all single-fp-source instructions now before you forget.  Seems like sqrt and convert, for sure.

coreyqh:
I added this coverpoint to fsqrt* and all variants of fcvt that round. All of these are still fully covered, hitting all of these new cross product bins.

PR #282, opened 2024-11-05
FLI immediate coverpoint

coreyqh:
* Created new coverpoint that checks the 5 bits of the rs1 field in the instruction binary, making sure each of the 32 values is exercised
* All FLI instructions are getting majority (87.5%) coverage, only missing hazard coverpoints
* FLI instructions should be fully covered once hazard refactoring is completed by the HMC team

### DIFF
--- a/bin/testgen.py
+++ b/bin/testgen.py
@@ -582,14 +582,12 @@ def make_fd(test, xlen, rng):
     desc = "cp_fd (Test destination fd = x" + str(r) + ")"
     writeCovVector(desc, rs1, rs2, r, rs1val, rs2val, immval, rdval, test, xlen, rs3=rs3, rs3val=rs3val)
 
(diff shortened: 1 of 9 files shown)


davidharrishmc:
This is a bit strange to cover this instruction with cp_fs1_fli.  It doesn't take a fs1.  Would it make sense to be cp_rs1_fli instead, or a different column?

coreyqh:
Ah, you're right. There had been an x incorrectly marking cp_fs1 before, so that is likely where this stemmed from. 

I'll change to cp_rs1_fli.

coreyqh:
Changed to cp_rs1_fli

davidharrishmc:
The function is still called make_fs1

coreyqh:
Oops, I missed that last night. 

These tests are now made with `make_rs1`

PR #290, opened 2024-11-07
Coverage of mulw 100% for RV64M

sherali9699:
The Testgen.py has been modified by adding a new condition in testgen in make rd corners functions for mulw instruction.


davidharrishmc:
This now has a merge conflict.  Can this be combined with the division coverage that was just accepted?

PR #294, opened 2024-11-07
Revert "Coverage improvement for RV32 R-Type Bit Manipulation"

davidharrishmc:
Reverts openhwgroup/cvw-arch-verif#288


davidharrishmc:
Hmm. The problem was before this PR.  Go ahead and issue this PR, but also fix the coverpoints so they run successfully with regression-wally --fcov.

PR #296, opened 2024-11-08
Enabled 'c' prefix for compressed instructions

Zain2050:
**PR#1077** in cvw enables **show_c_prefix**.

Removed code from covergroupgen.py and testgen.py which was specifically written for compressed instructions. 
With these changes, the coverage of all the compressed instructions dropped to zero. I made changes to the sample templates accordingly and brought the coverage back to 100%.

Previously some instructions such as c.ld and c.ldsp were both being shown as ld, and c.addi, c.addi16sp, c.addi4spn, c.nop were being represented as addi. With enabling 'c' prefix every instruction is now being represented by its own name, except for c.nop which is being displayed as c.addi but I added some special code in the sample template of Zca to handle c.nop.

### DIFF
--- a/bin/covergroupgen.py
+++ b/bin/covergroupgen.py
@@ -84,18 +84,10 @@ def customizeTemplate(covergroupTemplates, name, arch, instr):
         return ""
     instr_nodot = instr.replace(".", "_")
     template = template.replace("INSTRNODOT", instr_nodot)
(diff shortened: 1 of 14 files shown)


UmerShahidengr:
Wow. Wonderful work @Zain2050 . I thought it would take about a week, you did it in a few hours.

PR #299, opened 2024-11-09
Added NaNBoxing coverpoints

coreyqh:
* Currently limited to instructions that _always_ nanbox (All instructions that produce a half precision result, instructions that produce a single precision result that only exist with D enabled, etc) 
* Working on a way to get this coverage to all instructions that write a single precision value, if and only if D is supported
* All new bins hitting, so no lost coverage


davidharrishmc:
This is a start, but isn't going to catch any bugs.  It would be nice to have an assertion to check that the results aren't incorrectly NaN-boxed, but we don't have a systematic assertion flow yet.  I think you'll want to systematically check inputs that aren't correctly NaN-boxed for each source of each instruction that takes one or more fp sources.  Maybe a cp_rs*_corners_badnanbox with simple corners in the lower bits (0, 1, most positive, most negative, infinity, qnan, san), and random and zero bits in the upper bits?  This would ensure that badly NaN-boxed inputs are always detected, even on weird instructions like compares.

coreyqh:
I see. Unfortunately I think there will be instructions that need both cp_fs*_corners and bad NaNBox corners, so a new column will be needed. Are the coverpoints I made worth keeping or should I morph them into the bad NaNBox coverpoints?

davidharrishmc:
It’s harmless to keep them.

PR #302, opened 2024-11-12
ZicsrF Covergroup

coreyqh:


coreyqh:
* Tested with `regression-wally --fcov` without issue
* PR coming soon to CVW to add ZicsrF coverage to coverage.svh

davidharrishmc:
The walking test right now only tests writing 1s.  We might as well also test writing 0s.  The easy way to do this is add bins for csrrs and csrrc.  See the csrop and cp_mcsrwalk coverpoints in ZicsrM_coverage.svh.  

You can have walking 1s hit all XLEN bins rather than drop the bottom 8.  Our tests do that anyway, and it makes the code more regular. Update the test plan accordingly.

Dividing this file into three covergroups is getting repetitive.  How about reorganizing as a single covergroup and sharing code? Update the testplan accordingly.

The fflags_toggle bins are going to get hit by the CSR writes, even if we didn’t do any instructions that set CSRs.  Can you think about whether there’s a more robust way to check that the flags are actually set by executing instructions?

You’re moving fast.  I haven’t checked if the `ifdef XLEN64 business works yet.  Can you tell if it is working for you?

coreyqh:
* Condensed to one covergroup
* Changed walking ones to match the approach used for ZicsrM coverage
* Added `iff (ins.current.insn[6:0] != 7'b1110011)` to fflags_toggle coverpoint to ensure that something other than a CSR write causes the flags to be set

davidharrishmc:
This is graceful!

Huda’s recent PR means we should change all option.per_instance to 0.

mstatus_FS should be 2 bits (3 places)

We are missing a coverpoint in the testplan exercising FS transitioning between states.  Adding this will also let us replace mstatus_FS with mstatus_FS_n0 in the other coverpoint crosses, which will save a funch of sim time. Specifically, for each of the four values of mstatus_FS, we should do each type of operation that might cause it to change to dirty (or not): write to fcsr,  frm, fflags, do each type of floating-point instruction (including Zfa).  These operations should include fsw, which is floating-point but shouldn’t change FS.

coreyqh:
Added a coverpoint that crosses various instructions with the 4 possible values of mstatus.FS. This coverpoint covers illegal instruction traps when FS = 0, so all other coverpoints are modified to use non-zero FS.

davidharrishmc:
Let’s tab indent inside the `ifdef blocks for readability

Mstatus_FS_n0 has some meaningful bin names: init, clean, dirty

I’m not entirely clear on how mstatus.FS changes, but I think mstatus.FS will reflect the new value at the end of the instruction.  So I think mstatus_FS and mstuats_FS_n0 may need to be based on ins.prev.

Let’s add more instructions for cp_mstatus_FS_transition to check all the flavors touch FS.  Looks like at least fmadd, fsqrt, fsgnj, fmin, feq, flcass, fmv belong.  There are three types of converts, and each type should be checked separately because some touch FS and some don’t.  Might as well hit the Zfa instructions here too – ok if they fault.  Fli, fround, fmvh.

I updated mstatus.FS in the ZicsrF testplan.  Can you update csr_flags_set_m to match your coverpoint goal?  Also the _reserved coverpoint names no longer are really sensible because they walk every bit – can you rename?  In ZicsrM, it’s now cp_mcsrwalk.

https://docs.google.com/spreadsheets/d/1AJS_97x_23XMEzcvQugT8qqnwcWyC55VqSzpjBTMODM/edit?gid=29805079#gid=29805079

coreyqh:
* indented contents of `ifdefn blocks
* updated mstatus bin names
* Changed mstatus to check ins.prev.csr
* Added many instructions to cp_mstatus_FS_transition cross product
* overhauled cp_fflags_set_m to reflect testplan changes
  * now split into 5 different coverpoints for each fflag
* renamed walking ones coverpoints

PR #305, opened 2024-11-13
Initial ZicsrU covergroups

coreyqh:
* 2 of the 3 groups filled out
* passes regression without issue
* Will issue a PR to cvw including these covergroups in coverage.svh once this is merged


davidharrishmc:
The zeros and ones can be written as '0 and '1 in SystemVerilog rather than needing a bunch of cases based on XLEN.

I think you’ll want ucsr_cg to resemble mcsr_cg.  I think you can use the mcsr rs1_ones and rs1_corners.  If those don’t work, let’s change mcsr to match ucsr rather than doing the same idea two different ways in different files.  Be aware that mcsr is not getting full coverage yet, so there is something not right in either the coverpoints or the tests.

I think your cp_csrc and cp_csrs coverpoints are equivalent to mcsr cp_csrcs, but lets aim to do the same goal in the same way in both files.  Can you update the testplans to match?  There is going to be a lot of figuring out the most graceful way to express all these things.

In uprivinst, the coverpoint should be cp_uprivinst rather than cp_sprivinst.  I just fixed the testplan.

coreyqh:
* Refactored covergroups to be analogous to ZicsrM
  * Updated test plan to reflect these changes as well
* Added endianness covergroup

davidharrishmc:
In ucsr_cg, csrrs and csrrc are no longer used.

Uprivinst is ok draft for now but will need to be redone when we figure out how to handle nonretiring instructions.

Mstatus comments:
Per_instance = 0
Change option.comment (see Huda’s PR)

Move sd, ld, lwu into XLEN64 if, indent everything in there.
Use cp_doubleoffset on the sd/ld coverpoints

Let’s call mstatus_mpp bins U_Mode =2’b00 and M_Mode=2’b11 for consistency.

The mprv tests all need to run in priv_mode_m.  When MPRV=1, they should look at mstatus_ube rather than mstatus_mbe.

coreyqh:
* Tried to incorporate feedback and synchronize ZicsrU with ZicsrM as well as the testplans allow. 
* Removed option.comment and option.per_instance = 0 lines to propagate Huda's recent changes. 
* mret and sret use ins.prev.mode as with the ZicsrM covergroup.

PR #313, opened 2024-11-15
100% coverage of ZicsrM

davidharrishmc:
Exercises all 4096 registers
Looks at old values of mstatus/mode because sret/mret change them
Fixed bug in argument register for an ecall


jordancarlin:
@davidharrishmc is this deliberately deleting `templates/RV32I_coverage.svh`?

davidharrishmc:
Yes.  Obsolete.

PR #316, opened 2024-11-16
RV32 support for privileged coverage tests

davidharrishmc:


jordancarlin:
I can take a look at optimizing the Makefile when I have some time. Merging for now.

PR #321, opened 2024-11-18
Initial ExceptionM coverpoints

coreyqh:
* Work in progress, but passes regression
* need to think about how to check program counter among other things
* another thorough read of the priv spec will be necessary


coreyqh:
* Added the remaining coverpoints
* **MUST BE MERGED WITH ** openhwgroup/cvw#1118 **EITHER PR ON ITS OWN HANGS --FCOV**

coreyqh:
* Renamed to ExceptionsM to better match tests plan. 
* Once reviewed, I will reopen a PR into cvw to add ExceptionsM to fcov along with `define statements needed for the covergroups

davidharrishmc:
How about replacing the beq bin with branch, and make funct3 = ???.  Then the test will work for any branch.  How about cross product with the six funct3 flavors of branches? But it would affect the conditions for taken and not taken; they would need to be more general than rs1_rs2_eq/ne

cp_medeleg_ebreal is misspelled. Extra space after cross in this line.

medeleg is a WARL register, and I don’t think it holds all 1s because most of the columns don’t exist. So I think we will need a dot product of each exception with a walking 1 and walking 0 in the corresponding bit of medeleg to see that the exception is delegated or not.    More importantly, medeleg may not exist in a system without S mode, so I’ve moved that coverpoint to the ExceptionS testplan, and these should move there too.

Walking 0 probably should probably delegate most of the exceptions, but not ecall, or else the ecall handler will be running in supervisor mode and do funny stuff. I think our trap handler may not work gracefully when ecall is delegated to supervisor mode, and we will need to think carefully on this one.

Can you make an issue that the PMA covergroup should cover more illegal address bins than ExceptionM.  Ok to leave exception as is.

Misaligned_priority_load/store will need to look at only the upper bits of illegal_address, or it won’t be able to hit the cross products with the adr_LSBs.  Also, it should be considering adr_LSBs rather than pc_bit_1 and imm_bit_1.

coreyqh:
Added all 6 branches to the cross products. Fixed typo with ebreak, changed priority illegal address to look only at upper XLEN-3 bits and replaced pc_bit_1 and imm_bit_1 with adr_LSBs. 

I commented out the medeleg stuff. It will need overhaul when it is moved to ExceptionsS

davidharrishmc:
Bleu is misspelled twice
Branches_nottaken needs == rather than !=
For illegal_address_priority, the address needs to have 3 zeros appended to still be XLEN in size.
For illegal_address_prority, write XLEN-1:3 rather than use the ifdef.
The syntax is getting to be pretty sophisticated and you are using constructs that I don’t think we have tested.  They all look reasonable, but I’d feel better if we actually tested them.  Can you leave some comments for coverpoints that you’d like to have tested by a human when we have some assembly language tests?
I think it’s ok to deleted the commented out rs1_rs2_eq/ne

coreyqh:
Left some comments on complicated coverpoints. Let me know if I missed any that seem suspect.

PR #331, opened 2024-11-21
ZicntrM covergroup

coreyqh:
* WIP: missing all coverpoints that are specific to M mode
* Worked through with Roman, Madeleine, Marina, and Leilani
  * Roman and Madeleine working on ZicntrU
  * Marina and Leilani working on ZicntrS
  * PRs from these teams coming soon, pending review of this general structure and ability to push to github
* Coverpoints for writing mcounteren/scounteren decoupled from coverpoints for reading corresponding counters
  * Bit swizzling used to create coverpoints for reading counters with their enable bit set
  * Could be recombined using `prev` and `current` for back-to-back writes and reads. Open to feedback on whether this is more or less elegant/robust
  * Unsure of whether the walking ones should be filled with zeros or wildcards. For reading counters, it might make more sense not to care about other counters' enables


coreyqh:
All presently implemented coverpoints migrated to ZicntrU and ZicntrS.

davidharrishmc:
Cp_mcounteren_access_write and cp_scounteren_access_write are missing priv_mode_m in the cross

What are cp_*coutneren_access_read for?  I don’t see these in the testplan.  Can they be deleted?

I don’t see the coverpoints described in the testplan.

coreyqh:
Rewritten to reflect test plan changes.

davidharrishmc:
All of the coverpoints should have privmodem in the cross.
In csrr, should rs1 = 0 to just read without setting?  Does this affect any other coverage files?
Drop mtime = C01 from counters. There is no mtime CSR in a hart. Mtime is shared across all harts and is controlled by writign BFF8 offset in the CLINT.  C01 is the time CSR, which is a read-only shadow of mtime. 
Cp_cntr_write is missing mcountinhibit given in the testplan.
I propose deleting cp_inhibited_write and _read. Mcounthinibit could be unimplemented and all 0s, so we would be unable to hit the counters_inhibited coverpoint.  We are (should be) already exercising mcounterinhibit in cp_cntr_write, which will tell us if cycle or instret are not incrementing.

davidharrishmc:
counters_inhibited is unused but still in the file.  Drop it?

If you don’t drop counters_inhibited, mcounthinibit bit 1 is read-only zero, so the time_inhibited coverpoint is unreachable and should be removed.

coreyqh:
Whoops, I should have caught that. 

I removed another unused helper coverpoint as well.

PR #332, opened 2024-11-21
Fully covered Zbs extension

Daniyal-R-A:


davidharrishmc:
In make_uimm, can you just do for r in range(xlen) and drop all the rng stuff?  Can you make this change, and I can merge the PR?

It looks like you added cmp_rd_rs1 to the Zbs I instructions.  That’s an appropriate coverpoint, but I’m not sure why you only added it to those instructions.  The coverpoint seems like it would be applicable to Zbs R instructions as well.  The cmp_rd_rs1 and cmp_rd_rs1_rs2 would be applicable to Zbs R too, and so would corners and many other coverpoints.  Could you explain your though process.  See also Issue 300.

Daniyal-R-A:
Oh alright, I initially thought that the cover point made already were sufficient. But after seeing, issue 300, I am realising that its not the case. I will enable all further necessary coverpoints for I and R type as well. Thank you for letting me know. 

However, I dont see use of testing any corners in Zbs I type instructions as I exhaustively tested all uimm, rd and rs1 values. Please let me know if I am missing something.

davidharrishmc:
You’re testing all 32/64 uimm values.  You are testing all 32 rd and rs1 registers, but definitely not all 2^{32/64} rs1 values.  The rs1corners coverpoint will try some interesting values, rather than only testing with random values.

PR #338, opened 2024-11-24
ExceptionsF covergroup

coreyqh:


davidharrishmc:
Fmv_x_f and fmv_f_x have the wrong funct7 entries and don’t have rs2=funct3=0. This probably affects ZicsrF too.
Fclass, sqrt are missing 00000 in rs2.  Not a real bug, but might as well clean up.
Fli has the wrong funct7 and is missing funct3 = 000
Is cp_mstatus_fs_csr_write redundant with cp_mstatus_fs_illegal_instr because the csr operations are already in instrs?
The bug rate here is fairly high.  My review is going to be imperfect.  Let’s try to get another set of eyes on these privileged covergroups to reduce the bugs.  Maybe explaining them to a teammate or volunteer?  Or schedule a privileged review with Huda or Umer?

coreyqh:
I am meeting with some of the volunteers tomorrow. I'll have them do a proofread of my rough work. After all of the obvious mistakes are weeded out, I'll reach out to Huda or Umer for more rigorous feedback.

davidharrishmc:
Please add another coverpoint to ExceptionsF:

coreyqh:
* Corrected incorrect instructions in the instrs coverpoint
* removed cp_mstatus_fs_csr_write since cp_mstatus_fs_illegal_instr covers csrrw, csrrs, csrrc to the 3 FP csrs
* added cp_badfrm coverpoint, crosses a subset of instrs that have a frm field set to 111 (dyn) with 3 reserved csr frm values and non-zero mstatus.FS to ensure that these instructions are only illegal due to the illegal rounding mode

davidharrishmc:
Looks like fli still has wrong funct7 and funct3

PR #340, opened 2024-11-24
Created ExceptionsU covergroup

coreyqh:
Almost identical to ExceptionsM with priv_mode_m changed for priv_mode_u, plus mstatus.sie added to cp_mstatus_ie


davidharrishmc:
Cp_mstatus_ie should make ecalls from user mode to both supervisor and machine mode.  This involves crossing with medeleg[8], and priv_mode_u.

For completeness, cp_mstatus_ie in ExceptionsM should cross with priv_mode_m so we are sure to test an ecall from machine mode.  Medeleg[11] is a read-only zero so there is no need to cross it.

PR #341, opened 2024-11-25
ZicntrU covergroup

C0d3ing:
Co-authored with @madeleieine


davidharrishmc:
Looking good!

Scounteren might not exist if there is no S mode, so mcounteren_access_u cannot depend on it being 1s. Let’s remove scounteren_ones from ZicntrU and rely on ZicsrS points for that.  I’ve clarified the coverpoint in the spreadsheet.

Cp_mcounteren_write should be crossed with priv_mode_m. 

Ready to merge after these two changes.

davidharrishmc:
counters_mcounteren is missing cycleh_enabled, time_enabled, and instreth_enabled

PR #342, opened 2024-11-25
ZicntrS covergroup

Marina-Bellido:
Co-authored with @lanilei


davidharrishmc:
Please review this covergroup with somebody before reading my comments below.  I think you and your reviewer will catch these.

cp_mcounteren_write doesn’t specify which counter or privilege mode.  However, it’s already covered in ZicntrU, so you can drop it.
Cp_scounteren needss a cross with scounteren and priv_mode_s

Marina-Bellido:
Done and made new pr. Everything should be good now

PR #344, opened 2024-11-26
Improved Zicsr coverage

ahlyssa:
- all Zicsr instructions at 92.50% 
- missing hazards, will talk to Hamza to fix 
- added coverpoint (cp_imm5_corners)


davidharrishmc:
For the csrr*i tests, cp_imm_ones_zeros_5 and cp_imm_zero don’t hit anything not in cp_imm5_corners.  Suggest deleting them.  

Make_csr_toggle has a whole bunch of issues. You are calling csrw mscratch, x{rd}.  rd is a source register, not a destination.  You are using the csrw instruction, even though you are attempting to test many different instructions.  Therefore, this can’t possibly be giving you the coverage you intend.  Csr*I instructions can’t possibly toggle the upper bits, and crsrc clears bits instead of setting them.  I would suggest dropping the cp_rd_toggle coverpoint, and removing the make_csr_toggle function. 

What are you hoping to learn from cp_rd_sign and cp_rs1_sign coverpoints?  Can you drop them?

Testgen.py line 596 writes 0x000 as the immediate for csritype instructions.  These immediates should be 5 bits, not 12.  Suggest writing a nonzero decimal constant so the instruction actually does something.

PR #347, opened 2024-11-27
Zbb 32/64gc rori and roriw fully covered

Daniyal-R-A:


davidharrishmc:
ibtype_unsignedImm is getting messy.  Could you change the iddle part to imm = imm % xlen, rather than the if-else?

There is already a  cp_imm_shift_w coverpoint defined that does the same as RV64_cp_uimm_roriw.  Can you remove the cp_uimm column and RV64_cp_uimm_roriw coverpoint, and just put a w in the cp_imm_shift column?

Looks like rori in rv32 and rv64 testplans can have an x in the cp_imm_shift column, and then they don’t need a cp_uimm column either.

Daniyal-R-A:
Done!

PR #351, opened 2024-11-29
ExceptionsZalrsc Covergroup

coreyqh:
Unsure what to do about "cross match lr address" for sc from the test plan. I had trouble wrapping my head around the spec for these instructions.


davidharrishmc:
Let's cross with rd being 0 and 1, indicating the sc succeeded or not.

coreyqh:
Added rd boolean bins into the crosses for the two sc coverpoints

PR #352, opened 2024-11-29
ExceptionsZaamo Covergroup

coreyqh:


davidharrishmc:
This looks good but lets throw in all the other amo instructions for completeness while we are here.

coreyqh:
Added amoadd, amoand, amoor, amoxor, amomin, amominu, amomax, and amomaxu

PR #353, opened 2024-11-29
ExceptionsZicboU covergroup

coreyqh:


davidharrishmc:
Cbie=10 is reserved and will be unreachable, so you’ll need to exclude it (or just list the other 3)

coreyqh:
Made an `ignore_bins` for cbie = 2'b10

PR #354, opened 2024-11-29
Update Makefile priv headers

jordancarlin:


davidharrishmc:
Looks good, but Makefiles are not my strength.  Can you add ExceptionInstr-Tests.h and ExceptionInstrCompressed-Tests.h?  Both depend on illegalinstrtests.py.

jordancarlin:
Done

PR #357, opened 2024-11-30
ExceptionsZicboS Covergroup

coreyqh:


davidharrishmc:
The senvcfg tests didn't properly specify that menvcfg must be enabled. Seeing these coverpoints gives an idea of how to do this more comprehensively with cross products.  I've updated the two test plans to reduce the number of coverpoints while increasing the coverage. I left the old coverpoints for reference, but delete them once you have the new ones.  Sorry for the churn.  Keep your eyes out for improvements like this as you develop the coverpoints.

coreyqh:
Updated ZicboU and ZicboS to reflect the testplan changes

davidharrishmc:
Isn’t S mode 01

coreyqh:
Oops, now fixed

PR #360, opened 2024-12-02
Removed RV32/64ZcaZiscr & suppressed Zcd warnings

Zain2050:


Zain2050:
regex in covergroupgen.py looks for coverpoints as a word and a colon separated by one space character. These instructions were being misinterpreted as coverpoints which were causing the warnings. Therefore, I added a tab to fix it.

PR #365, opened 2024-12-05
Modified Makefile to run SV32 tests

Zain2050:
After compilation of SV32 tests, they are placed in a very odd arrangement. Within the vm_sv32/src folder there are separate folders for each test, each having two folders (dut & ref) within which elf files are placed. It becomes hectic to run the tests directly. Therefore, I added this code in makefile to make it easier. 
**make get_vm** looks within all the dut folders within vm_sv32/src and grabs **elf** files. They are then placed in a file **vm_tests.sh** in the form of 
```
wsim rv32gc /home/mzain/cvw/tests/riscof/work/riscv-arch-test/rv32i_m/vm_sv32/src/mstatus_tvm_test.S/dut/my.elf --fcov
wsim rv32gc /home/mzain/cvw/tests/riscof/work/riscv-arch-test/rv32i_m/vm_sv32/src/pmp_check_on_pa_S_mode.S/dut/my.elf --fcov
wsim rv32gc /home/mzain/cvw/tests/riscof/work/riscv-arch-test/rv32i_m/vm_sv32/src/pmp_check_on_pa_U_mode.S/dut/my.elf --fcov
```
Here we can comment out the tests which we don't want to run.

Then running **make vm** executes this file and fetches functional coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,22 @@ $(PRIVDIR)/Zicsr-CSR-Tests.h: bin/csrtests.py
 $(PRIVDIR)/ExceptionInstr-Tests.h $(PRIVDIR)/ExceptionInstrCompressed-Tests.h: bin/illegalinstrtests.py
 	bin/illegalinstrtests.py
 
+# This code is added especially for running VM SV32 tests
+# Replace --fcov with --lockstepverbose for debugging
+SV32DIR := ${WALLY}/tests/riscof/work/riscv-arch-test/rv32i_m/vm_sv32/src
+SV32OBJ = $(shell find $(SV32DIR)/*/dut -type f -name "*.$(OBJEXT)" | sort)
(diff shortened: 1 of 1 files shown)


jordancarlin:
It should be possible to run all of them using `regression-wally --fcov` by adding a single line to that script. Is the rationale for this just for debugging purposes so you can run only the sv32 tests? If so I wonder if we'd be better off updating iterelf to allow functional coverage. It should just be a matter of copying over some of the code from regression-wally.

Zain2050:
I just added this temporarily as SV32 tests and covergroups are in the debugging phase. I thought it would be more efficient and time-saving to just run the tests we are working on rather than all of them. 
In addition, vm_reserved tests are not working yet; it throws up some errors. Therefore, better not run them before fixing them.

PR #366, opened 2024-12-05
Critical typo fixes

coreyqh:
Necessary syntax error fixes. prevent wsim rv64gc from failing. Not sure why this was working for rv32 and not 64.


coreyqh:
I haven't pushed changes to the coverage.svh files in cvw, which is why this has gone unnoticed. Not as critical as I thought...

PR #410, opened 2025-01-30
zicntr tests and ZicntrM 

ahlyssa:
- created the zicntr.py to create the cntr tests 
- added registers for ZicntrM
- created walk files for ZicntrM, ZicntrU, and ZicntrU


jordancarlin:
This seems to be testing a lot of CSRs that are not part of Zicntr (like `fflags`, `mip`, and `satp`) and missing several of the CSRs that are actually part of the Zicntr testplan (all of the mhpmevent CSRs for example).

Also, you modified the existing Zicsr-*-Walk.S files instead of creating Zicntr-*-Walk.S files, which is what I think you intended based on the PR message.

davidharrishmc:
Ahlyssa, please review with Jordan and update.  Let me know when you are ready for my review.

PR #411, opened 2025-01-30
Added F ext. instrs. to D testplan for NaNBox CPs

coreyqh:
* Replicates F extension instructions in the D test plan to check NaNBoxing functionality, as these instructions only NaNBox when D is supported. 
* cp_NaNBox ensures results are properly NaNBoxed, and cp_badNB checks behavior when an argument is not properly NaNBoxed
* Near full coverage of all duplicated instructions. One coverage hole on `fsw` due to cp_badNB only exercising fs1. I'll raise an issue and circle back to this soon. 
* Added a trivial `make_nanbox` function to testgen, as instructions with only cp_nanbox do not have any tests made and miss all coverpoints. It simply writes a single random test.  
* Unfortunately other instructions need these coverpoints, but we cannot presently guarantee that D is supported with the current test plan structure. For example, instructions in the ZfaF test plan can NaNBox if D is supported. Possibly we need a ZfaZfhD test plan solely for these coverpoints? Same thing Zfhmin.


davidharrishmc:
Avoid asmcount coverpoints for the F instructions in D.
Looks like it is necessary to add ZfaZfhD and ZfhminD plans.

coreyqh:
* Modified testgen to create tests without relying on cp_asm_count
* Added ZfaZfhD and ZfhminD testplans 
* Closed coverage holes noted previously

davidharrishmc:
I'm pretty lost on this one.  Let's review in person.

PR #415, opened 2025-02-01
fixed init lib path

AnonymousVikram:
Realized that after the privileged tests were restructured, the path to `WALLY-init-lib` was broken. It'd be a good idea to come up with a way to parameterize this value, but in the meantime, this is a fix to allow the tests to run again


jordancarlin:
We should be able to use an include directive to GCC in the Makefile. I'll merge for now and then open another PR soon that updates the Makefile to avoid issues with finding it in the future.

PR #416, opened 2025-02-01
Update include paths

jordancarlin:
Follow up to #415. The Makefile is updated to include the main `cvw-arch-verif/tests` directory as a search path for header files.


jordancarlin:
Also remove the generated priv headers that were checked into the repo at some point and have the generated files created in a separate `priv/headers` directory to avoid this in the future.

jordancarlin:
And fix Sail binary name after the recent upstream updates

PR #419, opened 2025-02-04
Crypto hazards 100% covered, InterruptsM tests started

HamzaYusufJamal:


davidharrishmc:
This would be good to divide into two PRs because the hazard coverage can be accepted now and the InterruptsM hasn't been tested yet.

In set/reset msip and cause_external_interrupt need la, rather than li
And some ret statements
Test the program before committing

davidharrishmc:
Hazard changes could be pulled now, but are mixed in with the InterruptM that will take more work.
Replace j finished with j done and remove the finished code.
The now and soon suffixes should only apply to timer interrupts.  External and software interrupts are generated immediately so they don’t need a suffix.
Have you stepped through the code and checked that it is doing what you intended?
I think the timer compare is reset in the interrupt handler, so you don’t need the reset_timer_compare.  If the timer compare weren’t reset, you’d get an infinite loop of interrupts as soon as you returned from the trap handler.
I think software and external interrupts will need to be reset in the trap handler as well for the same reason and you won’t need reset_msip.
Similar issue for ssi, sei
Why does set_msip and reset_msip for RV32 load and store the high word of MSIP?  There is only one bit to change, and it shouldn’t need xlen.

HamzaYusufJamal:
I've kept the resets because I think I'll still need them when testing the cases in which MIE is 0. If the test doesn't generate an interrupt, then the trap handler won't actually reset any of the relevant registers, so in those cases I'll have to do it myself.

PR #427, opened 2025-02-07
Coverpoint cp_sstatus_sbe_endianness_* fully covered

Marina-Bellido:
Completed and fully covered 1 of the 3 coverpoints for EndianS.

### DIFF
--- a/tests/lockstep/priv/EndianS.S
+++ b/tests/lockstep/priv/EndianS.S
@@ -12,10 +12,9 @@
 
 main:
 
-///////////////////////////////// CHANGE CODE LATER !!!!
+///////////////////////////////// 
 // Endianness testing
-// cp_mstatush_mbe_endianness  -> WHERE DOES THIS COME FROM??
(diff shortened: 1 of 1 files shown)


davidharrishmc:
You wrote: cp_mstatus_mbe_endianness -> WHERE DOES THIS COME FROM??  On RV32, the MBE field is in mstatush[5] because there is no mstatus[37] since XLEN = 32.  Does that help?

However, aren’t you using mstatus_sbe in this file?  Should you change all the comments to that, in bit [36] or [4]?

What is the purpose of saving sstatus at line 46?  Should that be mstatus because you are touching mstatus?

You chould share the setendianness code doing the change back to supervisor mode and return.

The coverpoint was called cp_sstatus_sbe_endianness_*.  However, I’d been mistaken: there is no sstatus.SBE, instead it is mstatus.SBE.  I’ve changed the name in the testplan spreadsheet.  Can you change it in the code and in the coverpoints?

PR #435, opened 2025-02-11
fixed infinite loops in InterruptsM tests

HamzaYusufJamal:


HamzaYusufJamal:
Machine-mode and Supervisor-mode external interrupts are now being raised and handled properly. Simulation runs as expected, but InterruptsM coverpoints haven't been merged yet so still no coverage measurements available.

davidharrishmc:
Please pull all of the platform-dependent code (using PLIC and CLINT) to their own functions within WALLY-init-lib and mark them such so it is easy to replace on a platform with a different way of generating eternal or software interrupts.  InterruptsM should call code in WALLY-init-lib to cause and release interrupts, so this code can be shared with the other Interrupt{S/U} files.  Cause_timer_interrupt_now, etc. can move to init-lib because you’ll need it in the other tests too.
The four supervisor interrupt triggers can move to InterruptsS because we don’t know in a M system if there are S interrupts.
Looks like there are still more coverpoints to write, such as cp_interrupts, to finish the InterruptsM testplan.
InterruptsM coverpoints should be about ready.  Please test on the coverpoints.

davidharrishmc:
Wally-init-lib: 
93/97: x1 should be called ra to be consistent with 186/190
91/95/150/153: comments are out of date
Are lines 132-142 duplicating the clearing of STIP, SSIP, SEIP?
308: I don’t understand how andi t1, t1, -2 sets the lowest bit for hart 0.  It seems like it would clear the lsb and leave the other bits untouched.  Couldn’t 307-309 be replaced with sw zero, 0(t0), or am I missing something?
344:-350: shouldn’t there be a sd t0, 0(t1) for RV64 and a sw t0, 04(t1); sw t0, 4(t0) for RV32?
Set_msip is in InterruptsM, while reset_msip is in WALLY-init-lib.  I think set_msip, cause_external_interrupt_*, cause_timer_interrupt_now belong all in one place, which is presumably WALLY-init-lib.  A big header comment should separate these out from the basic trap handler and explain that they are utility functions for the various Interrupts* tests.

InterruptsM.S
369: why is this sw instead of sd?
375, 377: replace with sw t2, 4(t3)

davidharrishmc:
Can you add a short comment here with a summary of the changes you made, and how they relate to the open comments above?

HamzaYusufJamal:
Line numbers have been shifted around somewhat, but all comments have been addressed.
- x1 is now called ra everywhere in WALLY-init-lib.h, and comments reflect the storing/retrieval of t0, t1, and ra
- duplicating the STIP/SEIP/SSIP clears was unintentional, those have been removed
- comment for line 308 was wrong, it was supposed to say "clear lowest bit for hart 0 without affecting other bits." The behavior you described was the intended behavior.
- mistake in reset_timer_compare has been fixed, now has conditional for RV64 and RV32
- Line 369 in InterruptsM.S was a typo, should have been sd.
- mtimecmph is now replaced with a 4 offset from mtimecmp in all locations where it appears.

HamzaYusufJamal:
I've also opened issue #509 to keep track of the hacky method I used to handle vectored interrupts.

davidharrishmc:
Can you make the following fixes soon and post when they are updated?  I’ll merge the PR, and then you can finish the remainder of your coverage and we’ll only have to review the new stuff (wfi?) in your next PR.

WALLY-init-lib.h
373: delete comment about ignoring MTIMECMP upper word
InterruptsM:
Cp_interrupts: needs some more comments to explain your process. Begin with describing the coverpoint you are hitting and how you do it. What are you using s0, s1, s2 for?  Is s0 indicating MIE?  
If you put mie in the inner loop after the jal cause_*_interrupts, you don’t need to go through all of this twice, and don’t need a s0 loop variable for mie.   Along the lines of
Csrci t6, mstatus, 8 # clear mstatus.MIE, preventing interrupts
Jal cause_timer_interrupt_now # set mip.MTIP; shouldn’t do anything because interrupts are disabled
Csrsi t6, mstatus, 8 # set mstatus.MIE, allowing machine timer interrupt to fire if mie.MTIE=1
Jal reset_timer_compare # turn off timer interrupt if necessary
171: seems unnecessary to initialize s2 because you do it at 189 as well
177: This disrupts the other bits in mstatus, which can be important.  Never do a csrrw to mstatus; instead do a csrrc or csrrs of the bit you are trying to change.
178: there’s no need to shift s0 back if you’d changed 176 to slli t0, s0, 3.  Better yet, check s0 and either csrrsi t6, mstatus, 8 or csrrci t6, mstatus, 8, and drop the shift and csrrw.
181: needs more explanation.  What register is controlling which bit you are setting?  How do you p
184-185: the preferred idiom is slli t3, s1, 2
183: This works, but it would be more common to start s1 at 2 at line 170, and decrement s1 at the end of the loop instead of the beginning of the loop.  Same with 193.
198-212: weird way to write a case statement, and hard to follow without comments
215: why intented?
Cp_vectored: similar comments
247-271: largely repeats 191-215.  Can you make this a shared subroutine call? 
274: again hard to read without comments and explanation of what the registers are for
See spreadsheet entry E18.  This test is supposed to check that the highest priority interrupt is taken.  Mie should initially be 0, and then all of the applicable enables should be turned on at once so the highest priority interrupt is taken.  Right now you turn on all the enable bits first, and then begin turning on interrupts, and the first enabled interrupt will be taken rather the highest priority one.
285: what is t0 = 0 for?  Seems like it is changed at 290 without being used
290-292: could be replaced with
Andi t0, s1, 4 # is bit 2 set?
Slli t0, t0, 9 # mie.MEIE = bit 2 of s1
Similar for the next bits
309-10: same trick of andi t0, s2, 4
333-4: could be replaced with csrrw t6, mie, zero # clear all interrupt enables to start fresh
Could you leave a comment at the end that cp_wfi is not yet implemented?

HamzaYusufJamal:
All comments have been addressed, which also brought cp_priority up to 100% covered. There was one mistake in the InterruptsM covergroup that I fixed: cp_vectored says that mie.MEIE/MTIE/MSIE should be all ones, but the coverpoint had it as walking ones instead.

davidharrishmc:
InteruptsM:
I’ve defined STORE and LOAD in WALLY-init-lib.  You can now replace

#ifdef __riscv_xlen
#if __riscv_xlen == 32
addi sp, sp, -4
sw ra, 0(sp)
#elif __riscv_xlen == 64
addi sp, sp, -8
sd ra, 0(sp)
#endif
#else
ERROR: __riscv_xlen not defined
#endif


With
Addi sp, sp, 8-
STORE ra, 0(sp)
And the same thing with LOAD to restore the stack pointer

205,211 + 239-243: could be replaced by csrrw t6, mie, zero
211-223: could replace with csrs t6, mie, -1 or a constant with these three enables set
254: mie.MTIE -> mie.MSIE
280: comment should be # write to enable certain machine interrupts, triggering highest priority pending interrupt

There's a merge conflict on Init-lib now.
Please get one more review from a team member because this is a complicated and important one, and when you are all satisfied, let's merge.

jordancarlin:
> InteruptsM: I’ve defined STORE and LOAD in WALLY-init-lib. You can now replace
> 
> #ifdef __riscv_xlen
#if __riscv_xlen == 32
addi sp, sp, -4
sw ra, 0(sp)
#elif __riscv_xlen == 64
addi sp, sp, -8
sd ra, 0(sp)
#endif
#else
ERROR: __riscv_xlen not defined
#endif

We might want to change those to be `SREG` and `LREG` to match what riscv-arch-test uses so it is easier to migrate later.

davidharrishmc:
Sounds good.  Hamza, would you change that everywhere those are used?

HamzaYusufJamal:
I messed up and made the changes on the wrong branch, so there's a bunch of extra commits not from me because of the rebase. All of the comments have been addressed though, and LREG/SREG are now used everywhere I could see that is applicable.

jordancarlin:
> I messed up and made the changes on the wrong branch, so there's a bunch of extra commits not from me because of the rebase. All of the comments have been addressed though, and LREG/SREG are now used everywhere I could see that is applicable.

Maybe try a `git rebase upstream/main`?

HamzaYusufJamal:
I tried that but it looks like the other commits are still there. I can look into it further if it's important, but as far as I can tell, all of the changes I made are still there.

davidharrishmc:
WALLY-init-lib:
456-482: it would be cleaner to move the la t0, MTIME and la t4, MTIMECMP out of the #ifdef because it is shared between the two branches.  465 will need to change to t4 to be consistent.
473: does this need to be a bgtu?  I haven’t thought closely – please convince yourself one way or another

InterruptsM.S
299-302: this could be li t0, 0x20000A # clear mstatus.TW, MIE, SIE
305-306 could be andi t0, t0, 4 # check bit 2.  Similar for 311-12, 317-18
326: You seem to like ble zero, s2, here and elsewhere.  Most people write this as bge s2, zero because they write while (s2 >= 0) rather than while (0 <= s2).  Similar comments elsewhere in your compares.

HamzaYusufJamal:
I think bgtu is the right one. If I happen to load all 1s from MTIME, then add 0x100 to it, the overflow would make it 0x0...0ff. all 1s !> 0x0....0ff in signed comparison, so with a bgt I'd skip the carry addition, but overflow did occur.

tests/WALLY-init-lib.h
```diff
@@ -44,19 +44,38 @@
...
     #else
-        #define LOAD ld
-        #define STORE sd
+        #define LREG ld
+        #define SREG sd
     #endif
 #else
     ERROR: __riscv_xlen not defined
 #endif
+#define CLINT_MTIME_ADDRESS 0x0200BFF8
```
jordancarlin, line 56: Duplicate on line 42

tests/WALLY-init-lib.h
```diff
@@ -44,19 +44,38 @@
...
 
 .section .text.init
 .global rvtest_entry_point
 
+#define CLINT_BASE_ADDR 0x02000000
+#define PLIC_BASE_ADDR 0x0C000000
+#define GPIO_BASE_ADDR 0x10060000
```
jordancarlin, line 63: Let's keep all of the implementation specific constants at the top so they are easy to find and change.

tests/WALLY-init-lib.h
```diff
@@ -44,19 +44,38 @@
...
+
+#define MTIME           (CLINT_BASE_ADDR + 0xBFF8)
```
jordancarlin, line 65: How is this different from `CLINT_MTIME_ADDRESS`? Can we just have this and avoid the need for `CLINT_MTIME_ADDRESS` to be defined entirely?

tests/WALLY-init-lib.h
```diff
@@ -81,17 +100,31 @@
...
+    SREG ra, -16(tp)
     csrr t0, mcause     # Check the cause
     csrr t1, mtval      # And the trap value
     bgez t0, exception  # if msb is clear, it is an exception
 
-interrupt:              # must be a timer interrupt
+interrupt:              # must be an interrupt
     li t0, -1           # set mtimecmp to biggest number so it doesnt interrupt again
     li t1, 0x02004000   # MTIMECMP in CLINT
     #ifdef __riscv_xlen
```
jordancarlin, line 130: Seems like this section could be simplified to do an `SREG` to set mtimecmp and then an extra store to mtimecmph if it is rv32. Similar to what is done for stimecmp below.

tests/WALLY-init-lib.h
```diff
@@ -113,21 +146,28 @@ interrupt:              # must be a timer interrupt
...
     #endif
+
+    jal reset_msip
+    jal reset_external_interrupts
     li t0, 32
-    csrc sip, t0        # clears stimer interrupt
+    csrc mip, t0       # reset mip.STIP
+    csrci mip, 2       # reset mip.SSIP
+    li t0, 512              # 1 in bit 9
+    csrc mip, t0       # reset mip.SEIP
```
jordancarlin, line 156: Why not do all of this clearing in one step? `csrc` clears all of the bits that are set in rs1, so you should be able to set bits all three bits with the `li` and then clear the whole csr with one `csrc`.

tests/WALLY-init-lib.h
```diff
@@ -170,15 +209,18 @@ trap_return:             # return from trap handler
...
+    LREG ra, -16(tp)
     csrrw tp, mscratch, tp   # restore tp
     mret                     # return from trap
 
 write_tohost:
     la t1, tohost
     li t0, 1            # 1 for success, 3 for failure
-    STORE t0, 0(t1)     # write success code to tohost   
+    SREG t0, 0(t1)     # write success code to tohost   
+    SREG t0, 0(t1)     # write success code to tohost
```
jordancarlin, line 223: Duplicate lines?

tests/WALLY-init-lib.h
```diff
@@ -247,6 +291,265 @@ uncompressedillegalinstructionreturn:            # return from trap handler.  Fa
...
+    li t0, -1               # all 1s
+    la t1, MTIMECMP
+    #ifdef __riscv_xlen
+        #if __riscv_xlen == 32
+            sw t0, 0(t1)
+            sw t0, 4(t1)
+        #elif __riscv_xlen == 64
+            sd t0, 0(t1)
+        #endif
+    #else
```
jordancarlin, line 344: This can be simplified in the same way as mtimecmp above

tests/WALLY-init-lib.h
```diff
@@ -247,6 +291,265 @@ uncompressedillegalinstructionreturn:            # return from trap handler.  Fa
...
+    la t0, INT_PRIORITY_3
+    li t1, 1
+    sw t1, 0(t0)
+
+    # enable interrupts from source 3 (GPIO) in M-mode
+    la t0, INT_EN_00
+    li t1, 0b1000
+    sw t1, 0(t0)
+
+    # clear all interrupt enables to make sure interrupt doesn't go off prematurely
```
jordancarlin, line 382: Maybe I'm missing something, but shouldn't you clear all of the interrupt enables before turning on interrupts from the source instead of after?

tests/WALLY-init-lib.h
```diff
@@ -247,6 +291,265 @@ uncompressedillegalinstructionreturn:            # return from trap handler.  Fa
...
+                write_MTIMECMP:
+                sw t3, 0(t4)          # MTIMECMP = MTIME + 0x100
+                sw t2, 4(t4)          
+        #endif
+    #else
+        ERROR: __riscv_xlen not defined
+    #endif
+    ret
+
+cross_interrupts_m_EP:
```
jordancarlin, line 489: Many of your functions would benefit from a one or two line comment clearly articulating what they do. Especially for ones where the name may not be as immediately obvious like this one.

tests/lockstep/priv/InterruptsM.S
```diff
@@ -0,0 +1,323 @@
...
+#define MTIMECMPH       (CLINT_BASE_ADDR + 0x4004)
+
+#define THRESHOLD_0     (PLIC_BASE_ADDR + 0x200000)
+#define THRESHOLD_1     (PLIC_BASE_ADDR + 0x201000)
+#define INT_PRIORITY_3  (PLIC_BASE_ADDR + 0x00000C)
+#define INT_EN_00       (PLIC_BASE_ADDR + 0x002000)
+#define INT_EN_10       (PLIC_BASE_ADDR + 0x002080)
+
+#define GPIO_OUTPUT_EN  (GPIO_BASE_ADDR + 0x08)
+#define GPIO_OUTPUT_VAL (GPIO_BASE_ADDR + 0x0C)
```
jordancarlin, line 33: All of these are already defined in WALLY-init-lib.h.

jordancarlin (changes requested):
I still need to get through the actual InterruptM.S file, but here is a first round of comments.

tests/WALLY-init-lib.h
HamzaYusufJamal: I realized the code here is a duplicate of my reset_mtimecmp function so I just replaced it with jal reset_mtimecmp.

tests/WALLY-init-lib.h
HamzaYusufJamal: Pretty sure I just got lucky since I'm the only one using GPIO for my tests. Thanks for catching that.

PR #440, opened 2025-02-13
Changed rs1_val to fs1_val for FPR coverpoints

### LINKED ISSUE #439: ZicsrF coverpoints use rs1 instead of fs1
Several coverpoints in ZicsrF are using `rs1_val` for things that seem like they should be `fs1_val`. See example below:

```
  fs1_one: coverpoint ins.current.rs1_val[31:0] {
      bins one = {32'h3f800000};
  }
```

coreyqh:
resolves #439


jordancarlin:
I think this will break things further because the sample function uses rs1_val right now. I can look further into it once the refactor is merged.

PR #424, opened 2025-02-06
Refactor coverage collection

jordancarlin:
- Move riscvISACOV into main repo
- Add instruction decoder
- Lots of updates to functional coverage collection classes and functions

PR #441, opened 2025-02-13
Changed integer casting to XLEN_INT in CP templates

### LINKED ISSUE #437: Review size of integers used in coverpoints
Many of our coverpoints explicitly cast the values of registers/immediates/PC to 32 bits by using `'int`. This is incorrect for RV64 implementations and may give inaccurate coverage measurements as a result. There is a macro defined `` `XLEN_INT`` that is the appropriate size (`int` for RV32 and `longint` for RV64) that should probably be used instead for many of these. A careful review of all relevant coverpoints is needed.

coreyqh:
* Changed several int'() casts to use XLEN_INT macro
* Passed regression with no lost coverage

resolves #437

### DIFF
--- a/templates/cp_imm_mul.txt
+++ b/templates/cp_imm_mul.txt
@@ -1,4 +1,4 @@
-    cp_imm_mul : coverpoint int'(ins.current.imm) iff (ins.trap == 0) {
+    cp_imm_mul : coverpoint `XLEN_INT'(ins.current.imm) iff (ins.trap == 0) {
         //Immediate Multiples of 4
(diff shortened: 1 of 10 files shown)


jordancarlin:
It makes sense that this doesn’t change coverage because until my PR is merged they are equivalent.

davidharrishmc:
Looks good to me. @jordancarlin can you merge if satisfied?

jordancarlin:
I'm going to wait until after #424 is merged to avoid conflicts with that PR (they'll be much easier to resolve here).

coreyqh:
Merge conflict resolved

PR #444, opened 2025-02-13
InterruptsM Covergroup

coreyqh:
* Work in progress InterruptsM covergroup, coauthored with @jordancarlin 
* Missing a helper coverpoint for an external interrupt occurring, which is required for two of the cross products
  * Huda suggests modifying Wally Tracer/riscvTraceData to access the ExtInt signal
* Also missing a mtimecmp value coverpoint


coreyqh:
All coverpoints except cp_sei3 are ready for initial review. cp_sei3 is missing a coverpoint for plic/eic causing an external interrupt.

davidharrishmc:
This is a tricky one.  We can't count on mip.S*IP being nonzero without supervisor mode, so we need to remove a good number of coverpoints that I incorrectly put in the spreadsheet.  There are also quite a few implementation bugs.

Please review this one with another team member closely before reading my review below, in the hopes you find some additional bugs that I overlooked.

In mie_ones, there need to be question marks in the supervisor bins because these bits are wired to 0 in systems without supervisor mode. 

Sort the coverpoints in lines 30-65 to keep them organized by register in sequential order of bits.

Is there a reason the bit fields in lines 67-88 were in order 7,3,11 instead of 11,7,3 sorted?

Should mtvec_mode_zero be named mtvec_direct?

In clint_msip_set, you want rs2_val[0], rather than rs1_val[0]

In cp_trigger_mti, mip_ssip_one should be mip_mtip_one

In cp_trigger_msi, I think we can drop clint_msip_set.  The important thing to trigger the MSI is that mip.MSIP = 1.  The clint is a means to an end, just like MTIMECMP is a means to an end in cp_trigger_mti.

For cp_trigger_sti, we can’t ensure mip.STIP = 1 because it might be wired to 0 in a ssytem without supvervisor.  All we can do is test whether we can attempt to write a 1, but even that isn’t really useful because we can’t ensure mip.STIE = 1.  This feature is exercised in InterruptsS cp_trigger_changingtom_sti so I propose we delete it here.  Same applies to cp_trigger_ssi_mip, cp_trigger_sei_plic, cp_trigger_sei_sie.

InterruptsS is missing a coverpoint analogous to InterruptsM cp_trigger_sei_sie (which probably should be named cp_trigger_sei_mip for consistency with cp_trigger_ssi_mip) to trigger mip.SEIP by writing a 1 to mip.  Can you add it to the spreadsheet and coverplan?

For cp_global_ie, mstatus.SIE probably is read-only zero if supervisor mode is not supported, so it should be dropped from the coverpoint and spreadsheet.
https://github.com/riscv/riscv-isa-manual/pull/1864
I’ve added it to line 56 of interruptsS spreadsheet and trimmed SIE from line 15 of InterruptsM.

cp_vectored specified 2 x 3 bins, while the SystemVerilog coverpoint has 2 x 3 x 3 bins.  Given that mtvec.mode=00 is covered thoroughly by cp_interrupts, I’ve dropped that case from cp_vectored, getting it down to just 3 bins.  I made the same change in InterruptsS cp_vectored

I’ve removed mstatus.SIE from cp_wfi.

Cp_wfi should switch mip_mtip_one to look at both 0 and 1, to capture the case MTIMECMP causing the interrupt now vs. in the future.

Drop cp_sei1:3 from InterruptsM because mip.SEIP is tied to 0 without supervisor mode and these are already covered in InterruptsS. Please update spreadsheet and covergroups.

fcov/priv/InterruptsM_coverage.svh
```diff
@@ -0,0 +1,154 @@
...
+
+covergroup InterruptsM_cg with function sample(ins_interruptsm_t ins);
+    option.per_instance = 0; 
+
+    // building blocks for the main coverpoints
+    mstatus_mie: coverpoint ins.current.csr[12'h300][3]  { }
+    mstatus_sie: coverpoint ins.current.csr[12'h300][1]  { }
+    mstatus_tw:  coverpoint ins.current.csr[12'h300][21] { }
+    mie_ones: coverpoint ins.current.csr[12'h304][15:0] {
+        wildcard bins ones = {16'b????1?1?1?1?1?1?}; // ones in every field that is not tied to zero
```
C0d3ing, line 34: Should bit 13 not be conditional depending on whether or not the "Sscofpmf" extension for Count Overflow is enabled? 
image

coreyqh: We have bit 13 as a wildcard since we cannot guarantee that we can possibly set that bit to 1. This raises another issue of the supervisor interrupt enable bits, which are conditional upon S mode support. However, it is legal to have M without S mode.

coreyqh:
Refactored to remove S mode dependent coverpoints and to use the m_ext_intr, m_timer_intr, and m_soft_intr signals recently added to the tracer. 

Right now, cp_trigger_* checks for both the m_*_intr signal as well as the mip.m*ip bit being one. Is this overly redundant? If not, would it be more valuable to check the positive edge transition of mip.m*ip instead?

davidharrishmc:
Its, it seems redundant to check both m_timer_intr and mip_mtip_one.  How about checking the interrupt, and planning an assertion that mtip is 1 when the timer interrupt is asserted? Same for other machine mode.

The cases have collapsed.  Cp_global_ie is now like cp_interrupts except for mtvec_direct.  Let’s drop cp_global_ie because cp_vectored already catches the other case.  Please remove from both the test plan and coverpoints.

Let's get this merged and get coverage on the interrupt tests.

fcov/priv/InterruptsM_coverage.svh
```diff
@@ -0,0 +1,128 @@
...
+    // m_timer_intr: coverpoint ins.current.m_timer_intr {
+    //     bins mti = {1};
+    // }
+    // m_soft_intr: coverpoint ins.current.m_soft_intr {
+    //     bins msi = {1};
+    // }
+
+    // main coverpoints
+
+    cp_trigger_mti:      cross priv_mode_m, mstatus_mie, mie_ones, mip_mtip_one;
```
jordancarlin, line 109: @davidharrishmc What's the rationale for requiring all of the m-mode interrupt bits of mie to be enabled instead of just mie.MTIE? Similar for all the other trigger coverpoints.

jordancarlin:
> It seems redundant to check both m_timer_intr and mip_mtip_one. How about checking the mip, and planning an assertion that mtip is 1 when the timer interrupt is asserted?

If the idea is that these coverpoints are used to verify that a set of tests has full coverage and then those tests can be run on an implementation without needing to collect coverage, won't relying on assertions mean that this is not tested when not collecting coverage?

fcov/priv/InterruptsM_coverage.svh
```diff
@@ -0,0 +1,127 @@
...
+`define COVER_INTERRUPTSM
+typedef RISCV_instruction #(ILEN, XLEN, FLEN, VLEN, NHART, RETIRE) ins_interruptsm_t;
+
+covergroup InterruptsM_cg with function sample(ins_interruptsm_t ins);
+    option.per_instance = 0; 
+
+    // building blocks for the main coverpoints
+    
+    mstatus_mie_one: coverpoint ins.current.csr[12'h300][3] {
+        bins zero = {1};
```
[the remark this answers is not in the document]
coreyqh: Whoops, likely a copy paste error.

davidharrishmc:
@jordancarlin if there is a nonstandard implementation or an implementation without proper RVVI support that raises mip.MTIP without m_timer_inter, the coverpoint will still be hit.  In any case, the test program will do whatever is necessary on that implementation to trigger a timer interrupt.  I think we're good on this.

jordancarlin:
Sounds good. We might be able to back out our custom interrupt addition to RVVI then.

fcov/priv/InterruptsM_coverage.svh
davidharrishmc: Somewhat arbitrary.  Partly to make it easy to write the test.  Partly to see if there are undesired interactions with other enabled interrupts firing.  But there is a cross-product later that should hit all combinations of pending and enable, so I don't think this choice really mattered.

jordancarlin: Seems like a good enough reason. I don't see a reason to change it.

PR #447, opened 2025-02-13
Removed potentially impossible PC bins

### LINKED ISSUE #445: Update ExceptionsM_coverage.svh Cover Points
Noticed an unusual number of bins while running --fcov for ExceptionsM.S

The ExceptionsM Covergroup file needs updating to match the new test plans.
At path: $(WALLY)/addins/cvw-arch-verif/fcov/priv/ExceptionsM_coverage.svh

Include:
bins  pc_bit_1= {1'b0};
bins  imm_bit_1= {1'b1}
(respectively)

Image

Image

coreyqh:
* The original auto bins conflicted the test plan and were impossible to hit if C is not supported

resolves #445

### DIFF
--- a/fcov/priv/ExceptionsM_coverage.svh
+++ b/fcov/priv/ExceptionsM_coverage.svh
@@ -114,8 +114,10 @@ covergroup ExceptionsM_exceptions_cg with function sample(ins_exceptionsm_t ins)
        bins M_mode = {2'b11};
     }
     pc_bit_1: coverpoint ins.current.pc_rdata[1] {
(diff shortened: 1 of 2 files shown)


davidharrishmc:
Why does   
imm_bit_1: coverpoint ins.current.imm[1] {
        bins one = {'1};
    }
need bins one?  If that were blank, we'd get imm[1] being 0 and 1.  0 is tested in most normal branches, so only the one case will require a dedicated test, but the coverage is slightly more general.  Am I missing something?

PR #450, opened 2025-02-16
Simplify coverage extension selection

jordancarlin:
If a coverage file requires multiple extensions it will be automatically selected if all of those extensions are enabled instead of needing to be manually enabled.


jordancarlin:
The privileged ones can probably also be simplified. I'll revisit that later once they are more complete though.

PR #451, opened 2025-02-16
Disable printing every instruction by default with fcov

jordancarlin:


davidharrishmc:
Should the wsim command have --lockstep?

I had thought about removing the printing a few days ago, but feel like it's very helpful at the development state we are in.  If you want a FCOV_VERBOSE flag, should we make it on by default for now?

jordancarlin:
> Should the wsim command have --lockstep?

Not sure what lockstep has to do with this?

> I had thought about removing the printing a few days ago, but feel like it's very helpful at the development state we are in. If you want a FCOV_VERBOSE flag, should we make it on by default for now?

The old version of fcov didn’t print all the instructions by default, so I assumed we didn’t need it. For the longer tests the files also become quite large during regression. This version has it off by default and if you want to see the instructions you need to define `FCOV_VERBOSE`, which wsim lets you do using `--define “+define+FCOV_VERBOSE”`.

jordancarlin:
> > Should the wsim command have --lockstep?
> 
> Not sure what lockstep has to do with this?

Oh! You meant for `make sim` in the Makefile. Added.

PR #460, opened 2025-02-17
InterruptsU Covergroup

coreyqh:


davidharrishmc:
This is great.  It unearthed some bugs in the testplan as well.

All of interruptsU is supposed to run in user mode, so they need a cross with the mode.

There is no mideleg in a system without supervisor mode, so I’ve moved the mideleg items from InterruptsU to InterruptsS cp_wfi/cp_wfi_timeout. Same with mstatus.SIE and mstatus.SEIP.  You can drop mideleg_*and mstatus_sie and mip.seip/mie_seie from the testplan. Same for cp_user_sei.

coreyqh:
Removed mstatus.SIE from all cross products as it could be read only zero in a machine without S mode. I highlighted the affected test plan entries. Perhaps the cross products with mstatus.SIE could move into InterruptsS? 

cp_user_* and cp_wfi now use the m_ext_intr, m_soft_intr, and m_timer_intr signals along with the corresponding MIE bits being one.

PR #465, opened 2025-02-18
InterruptsS Covergroup

coreyqh:
WIP

Struggling to wrap my head around the coverpoints that check changing to s mode and then writing status.SIE.


davidharrishmc:
Please see extensive feedback on InterruptsM and update this before reviewing.

coreyqh:
Completely reworked since last review, now includes U and M mode cross products. 

Only missing cp_sei6 and cp_sei7 as I do not know a good way to check "turn[ing] of PLIC.SEIP"

davidharrishmc:
We discussed PLIC.SEIP.
Merge conflict now.

This is finding a lot of spreadsheet bugs!

As with InterruptM, just check mip, not interrupt pins, except on seip

Per privileged spec p. 94, each of the supervisor interrupts may not be implemented, in which case its interrupt pending and interrupt enable bits are read only zeros.  Our coverpoints right now assume they are all implemented.  We at least need a comment saying this will need to be relaxed.

Cp_trigger_sei_seip is problematic because we can’t write to mip while in supervisor mode.  How about checking if mip.SEIP = 1 and s_ext_intr = 0?

Let’s add mideleg = 0 to cp_interrupts to remove ambiguity

Cp_vectored should involve stvec_vectored rather than mtvec_vectored

For mideleg_mie_eq, let’s just look at the bottom 12 bits of each register in case the upper bits have funny stuff.

For cp_wfi_timeout, mstatus_tw should be 1, not {0/1} and mie.mtie should be {0/1, not 1.

Wfi is subtle.  Our coverpoint of detecting wfi looks reasonable, but there’s a chance the sample function will do something weird because wfi often doesn’t retire.  Just keep an eye out on this one.

I think the 4 cp_trigger_s* need to migrate *** from machine to supervisor, but probably partially overlap things already in supervisor?

Cp_vectored_m says to keep all three mie bits = 1, while the covperpoint is written with mie_s_walking.
Cp_trigger_changetom doesn’t have any mode changes.  How about dropping the changetom name in both coverpoints and spreadsheet?

We’d better add mideleg = 0 to the cp_sei* coverpoints and spreadsheet.

In cp_sei{1/2}, I’m concerned about mip_seip_rise and write_mip_seip in the same coverpoint.  Not certain both edge-triggered things get detected concurrently.  I think we can drop mip_seip_rise because we have write_mip_seip in both.

In cp_sei{3}, drop mip_seip_rise because we have s_ext_intr

Should no_sei be renamed no_s_ext_intr for consistency with s_ext_intr?

I think you have a plan for cp_sei6/7

Cp_global_ie: mip_m_walking and mie_m_walking are getting crossed, for 2x2x3x3 bins.  I think we’re fine with just 2x2x3 (mip = mie)

Cp_user_*: don’t need the interrupt pins; just use mip.*p

coreyqh:
* Removed redundant external interrupt pin coverpoints from cross products in favor of mip.M*IP signals
* Removed illegal mip write from cp_trigger_sei_seip by checking for mip.SEIP while the s_ext_intr pin is low
* Updated mi{e/p}_{s/m}_walking to only look at the relevant 3 signals
* Renamed *_changingtom_* coverpoints here and in the testplan
* Added mideleg=0s to cp_sei* here and in the testplan
* Dropped mip_seip_rise from cp_sei 1-3
* Combined cp_sei 6 & 7 by crossing the s_ext_intr pin being 0/1 and mip_seip being 0/1
  * Renamed s_ext_intr and no_s_ext_intr to s_ext_intr_high and s_ext_intr_low, respectively, to have s_ext_intr cover both values of this pin.

davidharrishmc:
In mideleg_mie_eq, ins.current.csr[12’h303] should also have [11:0]
Merge conflict to resolve

After this, I'm satisfied.  I'm sure more bugs will show up while trying to get the tests running, but that's ok.

@jordancarlin are you satisfied?

Issue #470, opened 2025-02-20
Cover Group cp_instr_adr_misaligned_jalr in ExceptionsM_coverage.svh


C0d3ing:
Incorrect number of bins.

Test plans state that Every combination of rs1_val[1:0] and offset[1] must be exercised.

I recommend creating a new cover point since the "imm_bit_1" cover point does not exercise all possible combinations of offset[1].

possible solution adding:

    offset: coverpoint ins.current.imm[1] {
    }

    cp_instr_adr_misaligned_jalr:            cross jalr, rs1_1_0, offset, priv_mode_m;

jordancarlin:
@C0d3ing is this resolved now that #478 is merged?

C0d3ing:
> [@C0d3ing](https://github.com/C0d3ing) is this resolved now that [#478](https://github.com/openhwgroup/cvw-arch-verif/pull/478) is merged?

Yes, and I have tested it and it is fully covered now.

PR #478, opened 2025-02-20
Added missing bins and fixed syntax for ExceptionsM Cover Goup 

C0d3ing:
Corrected a cross-product from issue #470.

Corrected branch_taken and branch_not_taken covergroups


davidharrishmc:
Can you adjust the name of the PR to match its contents?

PR #479, opened 2025-02-21
Fixed random corners in tests and RV64

davidharrishmc:


jordancarlin (commented):
Overall looks good. Left a few comments. 

The other significant thing I noticed is that ZicsrS and U still test all of the custom CSRs. They need to be updated in the same way as ZicsrM.

jordancarlin:
Also, there are a few custom compressed instructions that need to be avoided in `ExceptionsZc_instr_cg`. Specifically with compressed shifts.

jordancarlin:
Merging for now. Follow up PR will be made to address issues.

PR #486, opened 2025-02-24
Full coverage of several instruction types with ACT coverage metrics

jordancarlin:
Updated coverpoints and testgen to achieve full coverage according to the riscv-arch-test coverage for the following groups of instructions:
- R type
- I type
- Unsigned register and immediate corners


davidharrishmc:
Does this cause a noticeable change in runtime?

jordancarlin:
We should review which of these new corners seem worthwhile, remove any that don't, and then merge this PR.

PR #490, opened 2025-02-24
ExceptionsM 100% Coverage

C0d3ing:
Everything except the Instruction Access Fault cp is tested for.


tests/lockstep/priv/ExceptionsM.S
```diff
@@ -0,0 +1,432 @@
...
+// ExceptionsM.S
+//
+// Written: Roman De Santos rdesantos@hmc.edu 6 February 2025
+//
+// Purpose: Functional coverage test for Exceptions in M mode
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+///////////////////////////////////////////
+
+#include "../../WALLY-init-lib.h" // This path needs to be updated if this file is relocated
```
jordancarlin, line 12: It should work without specifying the path.

```suggestion
#include "WALLY-init-lib.h"
```

davidharrishmc:
I noticed that for jalr, offset[1:0] and rs1_val[1:0] both should be able to take on all 4 values, for 4 x 4 = 16 cross products.  I updated the coverpoints with PR #492, and you’ll need to double up on the set of tests accordingly to test off and even offsets.

Why is t4 incremented at line 200, 265, 337, 379?  Aren’t you testing the 8 possible offsets anyway with load_loop?

Why are lwu and ld, sd  commented out at 224-225, 254-255, 286, 361-2, 400?

What are your ecalls at 421, 426 doing?  They expect a value in a0.  I think you want it to be li a0, 3 just before the ecall to change to machine mode (do nothing)

Any luck figuring out why the instr_access_fault is doing something weird?

You made second round of commits, but it’s unclear exactly what they resolved or whether you’ve changed the coverage. It’s good to be specific in the PR exactly what has changed and whether there is still missing coverage.

davidharrishmc:
@C0d3ing were you able to fix up instr_access_fault?

C0d3ing:
> @C0d3ing were you able to fix up instr_access_fault?

Working on that right now.

C0d3ing:
Updated trap handler to handle Instruction Access Faults as defined by the RISCV privilege spec.

cp_instr_adr_misaligned_jalr now tests for all combinations of rs1_val[1:0].

Uncommented temporary changes.

Loaded a0= 3 before ecalls.

100% coverage on ExceptionsM

Removed TODO comments from ExceptionsM Coverage file (tested sizzling)

Added ExceptionsU to coverage collection.

C0d3ing:
This trap handler works for Instruction Access Faults now but if you would like a neater version I could spend some more time on it in the future after I finish ExceptionsZc. Just let me know.

davidharrishmc:
Please make the minor changes from Jordan and my reviews, and then I'm satisfied to merge!

Issue #495, opened 2025-02-25
Error in Zicsr Coverpoints


AnonymousVikram:
Many of the coverpoints look at the current mode, however this causes valid tests to fail when exercising illegal instructions, as the trap handler switches into machine mode. @coreyqh mentioned he could look into this and update as needed

AnonymousVikram:
I've already updated this in zicsrS and zicsrU, but this might be an issue in other coverpoints that need to be reviewed

jordancarlin:
Known to be an issue for exceptions and (likely) interrupts as well.

jordancarlin:
Looking into the root cause for this and likely planning to update this so that `ins.current.prev` gives the mode the instruction started in instead of ended in.

PR #496, opened 2025-02-25
100% Coverage for ZicsrS and ZicsrU

AnonymousVikram:
- Updated csrtests to have configurable custom csrs skipped
- Propagated skipped csrs in coverpoints
- Fixed checking old priv mode as speicifed in #495 
- Created tests for ZicsrS and ZicsrU


bin/csrtests.py
```diff
@@ -68,14 +72,27 @@ def csrtests(pathname):
 # generate repetitive assembly language tests
 
 # writable registers to test with walking 1s and 0s
-mregs = ["mstatus", "mcause", "misa", "medeleg", "mideleg", "mie", "mtvec", "mcounteren", "mscratch", "mepc", "mtval", "mip", "menvcfg", "mstatush", "mseccfg", "mseccfgh", "menvcfgh", "0x312"] # 0x312 is medelegeh; RV64 compiler isn't accepting the name
-sregs = ["sstatus", "scause", "sie", "stvec", "scounteren", "senvcfg", "sscratch", "sepc", "stval", "sip", "satp", "0x120"] # 0x120 is scountinhibit
+mregs = ["mstatus", "mcause", "misa", "medeleg", "mideleg", "mie", "mtvec", "mcounteren", "mscratch", "mepc", "mtval", "mip", "menvcfg", "mstatush", "mseccfg", "mseccfgh"]
```
jordancarlin, line 75: Why are some of these CSRs being removed?

fcov/priv/ZicsrS_coverage.svh
```diff
@@ -67,13 +67,41 @@ covergroup ZicsrS_scsr_cg with function sample(ins_t ins);
         bins mach_custom3[] = {[12'hFC0:12'hFFF]};
     }
     csr_no_satp: coverpoint ins.current.insn[31:20]  {
```
jordancarlin, line 69: Instead of almost duplicating this fairly large coverpoint, what about using an explicit ignore_bins in the cross product to exclude SATP? Take a look at the virtual memory coverpoints for reference.

tests/lockstep/priv/ZicsrS.S
```diff
@@ -0,0 +1,456 @@
...
+    csrr t2, mip # read back mip
+    csrrw t6, sip, t1 # restore sip
+    li a0, 1
+    ecall
+
+
+/////////////////////////////////
+// cp_sret
+/////////////////////////////////
+// Sweep executing sret in supervisor mode across cross-product of spp, mprv, tsr, spie, sie
```
jordancarlin, line 334: MPRV is no longer in the coverpoint, so I think this code and comment need to be updated.

bin/csrtests.py
AnonymousVikram: Realized it was unintentional, added back

jordancarlin: Can you restore the comment as well? And is scountinhibit being removed deliberately?

fcov/priv/ZicsrS_coverage.svh
AnonymousVikram: Done.

tests/lockstep/priv/ZicsrS.S
AnonymousVikram: fixed

bin/csrtests.py
AnonymousVikram: Done.

jordancarlin (approved):

davidharrishmc:
ZicsrM:
Why did you have to special case satp with 256-371?  When in machine mode, what is the harm of toggling all the bits of satp, so long as you restore them at the end?
If you are able to restore satp tests in machine mode, that affects csttests as well.

ZicsrS:
Could you move cp_satp and cp_shadow above cp_scause_write so the coverpoints are in the same order in the .S as in the spreadsheet?
Cp_shadow: give more comments explaining what you are doing (writing mstatus in machine, reading in supervisor)  Also at 215, point out that you are saving mstatus.
Suggest deleting 230-231 and 234-235 so you stay in machine mode rather than switching twice.  Same elsewhere.
It would help to comment srettest and setmstatus about what they are doing and what the arguments are.
336: you’re saving sstatus in s0, but then you trash s0 in srettest at 373 and never restore sstatus.  It would be good to restore sstatus at the end of cp_sret.  Better yet, you’re also tweaking mstatus.tsr, so it would be better to save and restore mstatus.  Line 402 restoring sstatus doesn’t do what you want; s0 was trashed at 373.
Can you replace 392-394 with csrrs t6, mstatus, t1?

ZicsrU:
Maybe cp_satp can be in the regular points, not special case?  It’s not listed on the spreadsheet testplan.
At 26-27, you enter user mode.  I don’t think it’s necessary to repeat that at 41-42, 46-47, 66-67, 76-77, 84-85.  Not sure there’s a reason to return to machine mode at 99-100.
I don’t think there’s a reason for postsrettest label, but some header comments about cp_csrr/cp_csrrw_corners/cp_csrcs would be useful before the calls to ZicsrU_CSR_Tests.
46-47: is this a repeat of 41-42?
I don’t think ZicsrU-Walk is required.  It’s not in the ZicsrU spreadsheet.  The flag testing is in ZicsrF.

davidharrishmc:
ZicsrS
330, 370: no need to save/restore sstatus because mstatus is a superset
376: comment should define what variable are passed in which register.  I think ra is the return address to the block that called srrettest, and needs to be saved before we call setmstatus.  A0 contains the desired status bits: mstatus.TSR in bit 22, sstatus.SPP in 8, ssstatus.SPIE in 5, sstatus.SIE in 1  
384:csrrw t6, sstatus, a0:   sstatus has many bits.  Should we only be tweaking bits 8, 5, and 1?
389: same thing: set mstatus.TSR to value pased in a0 bit 22

ZicsrU:
Keep code at 33-35 at start of cp_uprivintt to test a switch from user to user.

PR #502, opened 2025-03-01
ZicntrM 99.9% coverage

davidharrishmc:
The CSC demo was today, so I wrote ZicntrM so we'd have something to show.

There is an odd bug causing ImperasDV to hang on clearing minstret in lockstep 
https://github.com/openhwgroup/cvw/issues/1304

minstret is commented out in csrtests.py until this is fixed.
The other coverpoints are hit.


fcov/priv/ZicntrM_coverage.svh
```diff
@@ -174,7 +178,111 @@ covergroup ZicntrM_mcounters_cg with function sample(ins_t ins);
     `endif
 
     `ifdef XLEN64
-        walking_ones_zeros: coverpoint ins.current.rs1_val[63:0] {
+        walking_ones: coverpoint ins.current.rs1_val[63:0] {
```
jordancarlin, line 181: Would be good to make it clear in the names that the `walking_ones` coverpoint is targeting rs1 and the `walking_ones_zeros` coverpoint targets rs2. Almost missed that when reading through this.

tests/lockstep/priv/ZicntrM.S
```diff
@@ -0,0 +1,89 @@
...
+
+// General notes:
+// Use csrrw/csrrs/csrrc t6, csr, rs1    when modifying a CSR to also check the old value.
+
+#include "WALLY-init-lib.h"
+
+// define store instruction
+#ifdef __riscv_xlen
+    #if __riscv_xlen == 32
+       #define STORE sw
```
jordancarlin, line 19: Probably makes sense for this to be in a more central file so it can be used in multiple places?

tests/lockstep/priv/ZicntrM.S
```diff
@@ -0,0 +1,89 @@
...
+    ERROR: __riscv_xlen not defined
+#endif
+
+main:
+
+    # set up fast trap handler for uncompressed illegal instructions
+    # because Zicsr tests will throw a lot of illegal instruction traps while attempting to access CSRs
+    # This cuts the instruction count and runtime by nearly 2x
+    # However, don't blindly use this handler in other privileged tests
+    la t0, trap_handler_fastuncompressedillegalinstr
```
jordancarlin, line 33: Should this be using the fast trap handler? If so, this comment needs to be updated.

tests/lockstep/priv/ZicntrM.S
```diff
@@ -0,0 +1,89 @@
...
+
+/////////////////////////////////
+// cp_mtime_write
+/////////////////////////////////
+
+    li t0, 0x80
+    csrci mstatus, 0x8   # Turn off mstatus.MIE global interrupt enable
+
+// Testing walking zeros and ones for mtime in CLINT
+    la a0, CLINT_MTIME_ADDRESS  # address if mtime register
```
jordancarlin, line 44:
```suggestion
    la a0, CLINT_MTIME_ADDRESS  # address of mtime register
```

jordancarlin (commented):
Overall, looks good. Left a few small comments.

Also thinking that it might make sense for csrtests.py to guard h CSRs with ifdefs in the file it generates. That would simplify the handwritten portion of the test and would avoid the issues with CSRs like `scountinhibit` that are currently present.

tests/lockstep/priv/ZicntrM.S
davidharrishmc: Moved to WALLY-init-lib.

tests/lockstep/priv/ZicntrM.S
davidharrishmc: done

davidharrishmc:
Not sure how this helps scountinhibit, but I like the #if for h registers.  
Should be ready to merge.

bin/csrtests.py
```diff
@@ -70,7 +77,8 @@ def csrtests(pathname, skipCsrs):
 # generate repetitive assembly language tests
 
 # writable registers to test with walking 1s and 0s
-mregs = ["mstatus", "mcause", "misa", "medeleg", "mideleg", "mie", "mtvec", "mcounteren", "mscratch", "mepc", "mtval", "mip", "menvcfg", "mstatush", "mseccfg", "mseccfgh", "menvcfgh", "0x312"] # 0x312 is medelegeh; RV64 compiler isn't accepting the name
+mregs = ["mstatus", "mcause", "misa", "medeleg", "mideleg", "mie", "mtvec", "mcounteren", "mscratch", "mepc", "mtval", "mip", "menvcfg", "mseccfg"] 
+mregsh = ["mstatush", "mseccfgh", "menvcfgh", "0x312" ]# 0x312 is medelegeh; RV64 compiler isn't accepting the name
```
jordancarlin, line 81: Can 312 be switched back to medelgh now that it should only be compiled for rv32?

PR #503, opened 2025-03-02
Coverage for ExceptionM.S(100%), ExceptionsU.S(100%), and ExceptionsZc (100%)

C0d3ing:
Updated and organized cp_instr_adr_misaligned_jalr to reach 100% ExceptionM.s 

Initial commit for ExceptionsZc.S (Commented out since it still needs debugging)


tests/lockstep/priv/ExceptionsM.S
```diff
@@ -89,111 +87,107 @@ main:
...
-    addi    t1, t1, 11      # rs1[1:0] = 11
-    jalr    t1, t1, 11
+    addi    t1, t1, 5      # rs1[1:0] = 01
+    jalr    t1, t1, 7
     nop
     .word   0x00010013
 
-    // For rs1[1:0] = 00 (base value = 8)
-    // test offset[1:0] = 10, odd multiple of 2, rs1[1:0] = 00
+    // test offset[1:0] = 10, odd multiple of 2, rs1[1:0] = 10
```
davidharrishmc, line 181: Should offset be 11?

tests/lockstep/priv/ExceptionsZc.S
```diff
@@ -0,0 +1,262 @@
...
+#     c.ldsp  t4, 0(sp)       // Compressed load doubleword using stack-pointer offset
+# #endif
+
+#     // Compressed floating-point loads (if floating-point and compressed FP are supported)
+#     c.flw   f1, 0(t0)       // Compressed floating-point load word
+#     c.flwsp f2, 0(sp)       // Compressed floating-point load word from sp
+
+# #if __riscv_xlen == 64
+#     c.fld   f3, 0(t0)       // Compressed floating-point load doubleword
+#     c.fldsp f4, 0(sp)       // Compressed floating-point load doubleword from sp
```
davidharrishmc, line 152: Compressed loads are part of Zcd.  However, I think it's fine to keep them in this file - they will just throw an Illegal Instruction if not supported.  No need for xlen 64 checks.

tests/lockstep/priv/ExceptionsZc.S
```diff
@@ -0,0 +1,262 @@
...
+#     c.lw    t1, 0(t0)       // Compressed load word
+#     c.lwsp  t2, 0(sp)       // Compressed load word using stack-pointer offset
+
+# #if __riscv_xlen == 64
+#     c.ld    t3, 0(t0)       // Compressed load doubleword (RV64)
+#     c.ldsp  t4, 0(sp)       // Compressed load doubleword using stack-pointer offset
+# #endif
+
+#     // Compressed floating-point loads (if floating-point and compressed FP are supported)
+#     c.flw   f1, 0(t0)       // Compressed floating-point load word
```
davidharrishmc, line 147: Compressed flow are part of Zcf.  See below.  Fine to leave them in this file.

tests/lockstep/priv/ExceptionsZc.S
```diff
@@ -0,0 +1,262 @@
...
+// Written: Roman De Santos rdesantos@hmc.edu 28 February 2025
+//
+// Purpose: Functional coverage test for Exceptions in M mode
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+///////////////////////////////////////////
+
+# #include "WALLY-init-lib.h"
+
+# main:
```
davidharrishmc, line 13: Looks like the entire file is commented out.  Are you sure you tested it?

tests/lockstep/priv/ExceptionsZc.S
```diff
@@ -0,0 +1,262 @@
...
+#             fmv.d.x f3, t0                # Move the value from t0 to floating-point register f0.
+#         #endif
+#     #endif
+# #endif
+
+
+
+#     # --- Compressed Store Instructions ---
+#     # Note: c.sb and c.sh are nonstandard in the base C extension.
+#     # Ensure your toolchain/target supports these if you wish to test them.
```
davidharrishmc, line 227: Build command just needs to include Zcb.  I think it's harmless to have them in here - they will just trap if not supported.

tests/lockstep/priv/ExceptionsZc.S
```diff
@@ -0,0 +1,262 @@
+///////////////////////////////////////////
+// ExceptionsM.S
```
jordancarlin, line 2: Header needs to be updated

C0d3ing: Thanks,  updated.

C0d3ing:
Updated ExceptionsZc_coverage.svh to match riscV spec sheet (Instructions involving compressed sp access needed to check sp value not rs1 value)

Updated Makefile to add zcb flag to compiler for ExceptionsZc test conditionally.
-Please let me know if I should also conditionally add the c flag or if it's okay to leave it as a permanent flag(worried about accidental compiler optimization of 32-bit words to 16-bit words)

Added ExceptionsZc test and coverage report showed 100% coverage for both rv32 and rv64

*Minor concern: regression-wally times out for me during the fma tests so I am not able to 100% verify that regression ran. But when I run make priv and make sim (with the correct file path) the ExceptionsZc.S tests compile and run correctly.

davidharrishmc:
Good catch on the C flag.  It has to be conditional.  Otherwise the compiler will insert compressed instructions on other tests that are not intended to use them.

C0d3ing:
Updated C flag to be conditional

davidharrishmc:
Makefile
87: why do we need dut instead of ref?  Most of our test flow relies on ref.
123: don’t include C.  Maybe C should be at 105 instead?
143: don’t push your wsim command, or everyone else will need to edit it back to whatever they were using.
ExceptionsZc_coverage:
The testplan is trying to test compressed floating-point loads and stores in cp_{load/store}_address_misaligned, cp_{load/store}_access_fault. These instructions should be illegal if F is not supported.  The best solution would be to make an ExceptionsZcF for these cases.  However, let’s try leaving it as is for now, and these instructions will throw an Illegal Instruction exception when attempted, and that may be ok.  If we have trouble later when testing an MC100 core, we can come back to this.
77: I don’t understand why x_wdata[2] contains the value of sp?  Could you put some $display statements into your code to confirm you’re getting the behavior you want.  If you prove it is right, add comments explaining how this works. If it isn’t right, not sure what is best.
ExceptionsZc.S
You could merge the load and stores into the same loop so you don’t have to loop twice.
Do you want to write 16 bytes of known value to scratch before you start so you are reading initialized values?  Or did init-lib already do this?
15-16: indented one more than the other lines.  Are these necessary, or does WALLY-init-lib set up the trap handler?
37: worth commenting that some DUTs won’t support all of these flavors of loads/stores, and that’s ok because they will just throw an Illegal Instruction exception.
58: it’s a bit strange to increment sp before the accesses rather than after.  Should this move down by 73 so the 3 lsbs are the same for normal and sp accesses?
122: here you are initializing sp to scratch, while at 27 it was pointing somewhere else.  Do you want to make them the same?
152: are you sure this is right? At 58, you were adding 1 not t0. See also 196.
196: what is this for?  T0 doesn’t seem initialized in this portion fo the code.

davidharrishmc:
ExceptionsZc
27: agreed that writing 16 bytes of unique stuff to scratch first would be worthwhile.
37: a2 doesn’t appear to be initialized before you use it.  Does 108 need to move up here?  I’m surprised the code is getting coverage – maybe it is trapping when a4 = 0, and then working for other a4 values?
76: attempts to sw a3, but a3 isn’t initialized to anything know.  Suggest giving a3 a value first.  Same with f8.

This is subtle stuff.  Can you get a review from another teammate as well?

C0d3ing:
ExceptionsU.S:
Added ExceptionsU (very similar to ExceptionsM) 96% coverage. Missing cp_mstatus_ie.

ExceptionsU_coverage:
Updated  cross products, cp_instr_adr_misaligned_branch_nottaken and  cp_instr_adr_misaligned_jalr, to match test plans.
Changed cp priv_mode_u  to use the previous privilege mode since it changes to machine mode when it traps.

davidharrishmc:
Turn off display in exceptionsu_sample

ExceptionsZc:
Looks like you made some changes since my last review but haven’t made all changes yet.  Here are some things that jump out.
34: 38 is now initializing a2, so 34 isn’t needed
27: write 16 bytes of unique stuff to scratch would be useful before trying to read it at 41-44
79: a3 doesn’t have a known value in it yet – initialize.  Same with f8.

ExceptionsU.S: 
It looks like you haven’t reviewed the lockstepverbose log closely yet.  This code presently is doing some things you didn’t intend. Please review and clean up any unintended behavior.
7 update comment to U mode
20-21: must move after 29 because mtvec and mstatus are not writable in u mode
Are you sure 24-25 is necessary? I think mtvec is written in WALLY-init-lib.  Maybe this affects other files?

355 says REMOVE AFTER TESTING
464-492: rethink this.  You can’t access the m registers in U mode

jordancarlin:
I'll look into the sp issue (x_wdata) in the next few days before we merge this. I think the only reason this change was necessary is because of a bug in the backend, and I'd rather fix the root issue instead of changing the coverpoint.

davidharrishmc:
How about we merge when everything else is satisfactory, and then we can change all the prev back to current if necessary.  It would be great to get this PR done and merged.

jordancarlin:
> How about we merge when everything else is satisfactory, and then we can change all the prev back to current if necessary. It would be great to get this PR done and merged.

This isn't an issue with prev vs current. It's about accessing sp when it is an implicit register argument.

davidharrishmc:
Got it.  Thanks.

C0d3ing:
Just realized I have some unintended behavior in ExceptionsU currently. Will fix that right now.

fcov/priv/ExceptionsU_coverage.svh
```diff
@@ -113,7 +110,8 @@ covergroup ExceptionsU_exceptions_cg with function sample(ins_t ins);
     mstatus_SIE: coverpoint ins.prev.csr[12'h300][1] {
         // auto fills 1 and 0
     }
-    medelegb8: coverpoint ins.current.csr[12'h302][8]{
+
+    medelegb8: coverpoint ins.current.csr[12'h302][7]{
```
jordancarlin, line 114: Why is this being changed? Bit 7 is Store/AMO access fault but this coverpoint is crossed with ecall later, which is bit 8.

fcov/priv/ExceptionsU_coverage.svh
```diff
@@ -156,4 +153,13 @@ endgroup
 
 function void exceptionsu_sample(int hart, int issue, ins_t ins);
     ExceptionsU_exceptions_cg.sample(ins);
+
+    //$display("Instruction is: PC %h: %h = %s (rd = %h rs1 = %h rs2 = %h) trap = %b mode = %b (old mode %b) mstatus %h (old mstatus %h).  Retired: %d",ins.current.pc_rdata, ins.current.insn, ins.current.disass, ins.current.rd_val, ins.current.rs1_val, ins.current.rs2_val, ins.current.trap, ins.current.mode, ins.prev.mode, ins.current.csr[12'h300], ins.prev.csr[12'h300], ins.current.csr[12'hB02]);
+    $display(" ecall: %b, MIEPrev: %b, SIEPrev: %b, medelegb8, CurPriv: %b, PrevPriv: %b",
```
jordancarlin, line 158: Remove or comment out display statements

C0d3ing:
ExceptionsU_coverage:
-Removed coverage
-Reverted medelegb8 bit value (left accidentally)

ExceptionsU:
-Added a rudimentary S mode trap handler since medeleg lowers trap privileges. This trap handler stores and updates the PC value, then raises privilege with an ecall. This means in u mode, ECalls for privilege changes can happen as normal.

ExceptionsZc_coverage:
Reverted changes to original functionality since rs1 is filled with the sp value.

On my end, coverage is 100% for all 3 Exceptions tests.

C0d3ing:
-Changed the ExceptionsU_coverage to match the updated testplans
-Updated ExceptionsU match updated testplan
-Implemented ExceptionZc changes

PR #505, opened 2025-03-02
Vector instructions added to disassembler

jacassidy:


davidharrishmc:
I think you accidentally committed all files.  You should only be making a PR for a few files that you modified.

You'll also need to sign the ECA.

PR #508, opened 2025-03-05
Interrupts Sstc Covergroup

coreyqh:


davidharrishmc:
Rename write_stimecmp because it is used in csr reads
Otherwise looks great!

PR #511, opened 2025-03-06
increased coverage in ZicntrU

ahlyssa:
- i had added what was included in the pr for 100% coverage of ZinctrM
- in csrtests there are two new functions mcounterenwalk and mcounterenwalku that write walking 1s and 0s to mcounteren  and then read from corresponding counter and counterh in U-mode and M-mode 
- I am currently running into this error when I run make "make: *** No rule to make target 'tests/lockstep/priv/rv32/EndianM.elf', needed by 'priv'.  Stop." so I am unsure of what the coverage is at but when I had first started the coverage was ~45%


davidharrishmc:
As we discussed, there's unfortunately a big merge conflict.  Please close this PR and sort out the merge, and then reissue when you have good ZicntrU coverage.

PR #513, opened 2025-03-06
Templates for Vector Registers

Isabella-Hottenrott:
Coverpoint Templates for Vector Register Assignments and Vector Register Contents


davidharrishmc:
I think we can drop vd_vs1_eqval, vd_vs1_vs2_eqval, vd_vs2_eqval.  This is testing that there is an identity element in one of the sources.  We found this fall that targeting a particular vd value was cumbersome and didn’t give much more confidence into the design.

Otherwise, these look good to me.  @jordancarlin any other feedback?

jordancarlin (commented):
I think a lot of this might have been based on an older version of the integer coverpoints. It should be quick to update, but can you go through and compare with the current integer versions? For example, you have
`cmp_vd_vs1 : coverpoint ins.current.insn[11:7]  iff (ins.current.vd == ins.current.vs1 & ins.trap == 0 )`

and the other version has 
`cmp_rd_rs1 : coverpoint ins.get_gpr_reg(ins.current.rd)  iff (ins.current.rd == ins.current.rs1 & ins.trap == 0 )`

jordancarlin:
Something seems to have gone wrong with a merge and you are undoing all of the other recent commits from other people.

PR #517, opened 2025-03-07
Disable vector generation

jordancarlin:
Functional coverage currently broken because of vector. This temporarily disables it.


jordancarlin:
Merging because this is blocking the team from making progress and it is a simple PR

PR #519, opened 2025-03-07
Add standard coverpoint file and update priv mode to use prev

jordancarlin:
Adds a new standard coverpoint file that can be included in other coverpoints to reduce duplication of basic coverpoints. As a proof of concept this is used for priv_mode coverpoints.

All priv mode coverpoints are also switched to use prev to get the privilege mode at the end of the previous instruction.


jordancarlin:
Most of the coverpoints were a simple switch, but the VM coverpoints were more involved and could do with a close review.

davidharrishmc:
I don't have time to review the VM points right now.  Instead of risking breaking them, do you want to just revert them?

Zain2050 (approved):
VM covergroups look good! Can be merged.

PR #532, opened 2025-03-19
RV32VM_coverage Coverpoints

C0d3ing:
When trying to run make sim there were coverpoint issues in rv32_priv/RV32VM_coverage.svh
Please advise if these fixes are correct since I am unfamiliar with page types.
After my changes `make sim` executes correctly.

Changes:
created bins giga under coverpoint PageType_i and PageType_d

Renamed cover points to have unique bin names from lines 707 to 729 in the change log

### DIFF
--- a/fcov/rv32_priv/RV32VM_coverage.svh
+++ b/fcov/rv32_priv/RV32VM_coverage.svh
@@ -170,12 +170,10 @@ covergroup RV32VM_mstatus_mprv_cg with function sample(ins_t ins);
     PageType_i: coverpoint ins.current.page_type_i {
         bins mega = {2'b01};
         bins kilo = {2'd0};
-        bins giga = {2'b10};
     }
     PageType_d: coverpoint ins.current.page_type_d {
         bins mega = {2'b01};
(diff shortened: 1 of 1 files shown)


Zain2050:
Yeah, we don't need to add gigapages to the coverpoints. The issue is that I made a typo by writing **ignore PageType_d.giga** instead of **ignore PageType_i.giga**. Thanks for pointing it out, totally my fault.
@C0d3ing Changes from line 710 downwards fix this, while you should remove the giga bins that you added.

jordancarlin:
Won't ignoring the giga bin fail if the giga bin doesn't exits? Seems like all of those ignore bins would need to be removed too.

Zain2050:
Absolutely! Didn't see that the added giga bins are in the PageType coverpoints. 
Yes, removing the ignore giga bins is the correct fix. I guess while working simultaneously on sv32 and sv39 I forgot that sv32 PageType coverpoints don't have giga and doesn't need to be ignored. Should have properly tested SV32 as well before making PR. My bad.

jordancarlin (approved):

PR #535, opened 2025-03-20
ExceptionsS.S (100% Coverage)

C0d3ing:
RISCV_coverage_config.svh:
-Included ExceptionsS.S

ExceptionsS_coverage.svh:
-Updated Coverpoints to match testplan
Note: I changed cp_xstatus_ie on the test plan since writing to medelg bit 9 causes an ecalls trap. 
I updated the test plan to read, "ecall cross {mode = user/supervisor, medeleg[8] ={0/1}, mie = {0/1}, sie = {0/1}" under the cover point column of cp_xstatus_ie.

Let me know if the change needs to be reverted. Previously the test plan read, "ecall cross {mode = user/supervisor, medeleg[9:8] four values, mie = {0/1}, sie = {0/1}"

ExceptionsM.S:
- Removed a redundant trap setup since it is already handled by WALLY-init-lib.h

ExceptionsS.S
-Assembly Tests for all coverpoints


fcov/priv/ExceptionsS_coverage.svh
```diff
@@ -122,6 +119,7 @@ covergroup ExceptionsS_exceptions_cg with function sample(ins_t ins);
         bins zero = {0};
     }
     imm_bit_1: coverpoint ins.current.imm[1] {
+        bins zero = {1};
```
jordancarlin, line 122: Should probably name this one instead of zero.

davidharrishmc:
The s_traphandler needs rethinking and explanation.  It does not end in a sret, which seems fundamentally wrong. There are no comments explaining its overall architecture.  It seems to make assumptions about values in ra and a3 that are not documented.  It trashes the contents of t6.  It doesn’t explain why it changes privilege modes.  If it is a minimal viable hack to get coverage running until the riscv-arch-test trap handler is substituted, that’s ok, but say so and explain what it is currently doing.
I’m having a lot of difficulty following the flow of the program and what each piece is trying to do. Could you take another pass through this and try to make it as simple and readable as possible?  
Here are some initial comments, but I can’t digest it all yet.
The trap handler belongs at the end of the file, not before main.
33-35: why is floating-point being enabled?
37: why is there a jump to cp_medeleg_done followed by > 400 lines of code?  Why not move the code between the jump and cp_medeleg_done to the end and remove the jump?  cp_medeleg_done doesn’t seem to relate to anything being done.
477: three modes -> three privilege modes
479: There is a long and somwhat random list here of coverpoints.  It’s unclear if these apply to just supervisor mode, or all of cp_medeleg_msu.  Some coverpoints such as cp_illegal_isntruction_seed are in this list but not required for cp_medeleg_msu.  Other coverpoints such as cp_instr_adr_misaligned_branch are not in this list.  Can you either apply all tests to all privilege modes with medeleg, or only apply the minimal set required for cp_medeleg_msu, or explain why you’ve picked certain ones and not others?
_sum is out of date; now _msu
Space between // and comment
510/518: “strap handler”
I don’t understand all of the //mtrap/strap loads from 594-766.  Is there a reason arguments have to be passed in both a0 and a3, or could the supervisor trap handler accept a0 as well? 
751: “load into machine mode” – do you mean “change to machine mode”?  “load” is a type of instruction.

C0d3ing:
WALLY-init-lib.h:

- To implement the s_traphandler in ExceptionsS.S that allowed for privilege mode changes to any privilege mode (not just U and S), I had to add new modes 5-9 to signal that these are ecalls from the s_traphandler. This makes sure that the m_traphandler uses sepc instead of mepc when called from s_traphandler. This ensures that the s_traphandler provides easy changes to privilege mode by loading the desired privilege into a0 before an ecall.
- I updated the m_traphandler to preserve the value of a0. This ensures that when there is a delegated exception the s_traphandler has a valid mode to return to. a0 can only be used to keep track of the privilege mode.
- jalr writes ra = PC+4 but since the PC counter update is handled by m_trap already I subtracted 4 from ra incase it was an Instruction Access fault (line 183) 

ExceptionsS.S:

- Removed enabling code for floating point
- Updated trap handler and moved to the end of program. Handles simple privilege changes
- Moved medeleg function to the end.
- Removed cover points that did not need to be tested with different medeleg values and placed them in the main function
- s_trap handler now uses a0 instead of a3 
- medeleg uses a0 to determine test mode
- Updated comments according to previous suggestions

C0d3ing:
ExceptionsS.S:

- Clarified how the s_traphandler interacts with the m_traphandler when medeleg[8] = 1 (user mode ecalls are delegated).

davidharrishmc:
WALLY-init-lib:
It is unclear whether these changes are temporary and intended to be deleted later.  It would 
40: notes other arguments are reserved, but doesn’t define what each of them do. I think you want to explain here that the supervisor trap handler in ExceptionsS.S sometimes makes ecalls to this trap handler, and what this handler does in response.
147: says temporary code, but doesn’t explain what it is for.  traphendler typo
183: I don’t understand how subtracting 4 always works.  It seems this should break the regular exception handler.
ExceptionsS.S
34: comment reverses order of overpoints from 39 and 44.  Be consistent
35: comment says a3 but code says a0.  Same at 40 and 45
60: comment is inconsistent with code
361: It’s unstated what traps this handler receives.  The notes only mention ecall from user-level, but I think this is invoked by all delegated traps from user or supervisor mode.  Right now, it looks like it always terminates with an ecall to machine mode (395) to change mode to what is in a0.  I thought that should only happen for ecalls from user mode.  I thought the other traps should do a sret.
363: How about “This may be replaced by the generic riscv-arch-test trap handler in the future to log trap signatures, but is adequate for lockstep testing.”
409-412: this talks about 17 walking 1s in the upper 32 bits of medelegh.  I’m not sure what this is for – medelegh is not mentioned in the testplan or coverpoints and not used in the assembly.  Also, it seems like this loop might run extra iterations after the medeleg 1011_0001_1111_1111.
418: comment says 19 but a2 = 20.  Inconsistent with 17 above.
423/428: why mv t0 to a3 and then a3 to a0?  If necessary, explain. Why double extra space?
789-790: it would be clearr to load a temporary with 9 and compare to a1 rather than do this subtraction.  Same at 793.
Please proofread.  There are still a lot of inconsistencies.

C0d3ing:
*Not ready for review*
still need to proof read comments.

Pushed updated traphandler that uses sret for other delegated causes. Only delegated ecalls are raised to machine mode trap handler for privilege changes.

C0d3ing:
ExceptionsS.S

- Updated traphandler to skip machine mode ecall if it is not a user mode ecall that has been delegated to the s_traphandler.
- Updated loop counters to compare to a value instead of subtracting and checking if it was zero
- Comment revisions
- Updated s_traphandler description for future readability
- Loop iterates the expected number of times (19 but skips over 9-11 to avoid ecalls loops and to reduce testing time)
- removed any mention of medelegh

WALLY-init-lib.h:

- Described the purpose of the reserved bits
- removed the temporary code blocks
- described why subtracting 4 works for Instruction Access faults

PR #536, opened 2025-03-21
ExceptionsF (100% Coverage)

C0d3ing:
RISCV_coverage_config.svh:

- Added ExceptionsF to coverage collection

ExceptionsF_coverage.svh:

- Updated ins.prev to ins.current
- Removed redundant coverpoint comment 

ExceptionsF.S:

- Added Exceptions F assembly code

### DIFF
--- a/fcov/priv/ExceptionsF_coverage.svh
+++ b/fcov/priv/ExceptionsF_coverage.svh
@@ -25,25 +25,17 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
     option.per_instance = 0; 
 
     // building blocks for the main coverpoints
-    mstatus_FS_zero: coverpoint ins.prev.csr[12'h300][14:13] {
+    mstatus_FS_zero: coverpoint ins.current.csr[12'h300][14:13] {
         bins disabled = {2'b00};
     }
-    mstatus_FS_nonzero: coverpoint ins.prev.csr[12'h300][14:13] {
-        bins enabled = {!2'b00};
+    mstatus_FS_nonzero: coverpoint ins.current.csr[12'h300][14:13] {
+        bins enabled = {2'b11};
     }
-    frm_illegal: coverpoint ins.prev.csr[12'h003][7:5] {
+    frm_illegal: coverpoint ins.current.csr[12'h003][7:5] {
         bins reserved_5 = {3'b101};
         bins reserved_6 = {3'b110};
         bins reserved_7 = {3'b111};
     }
-    // csrrw: coverpoint ins.current.insn {
-    //     wildcard bins csrrw = {32'b????????????_?????_001_?????_1110011}; 
-    // }
-    // fp_csrs: coverpoint ins.current.insn[31:20] {
-    //     bins fcsr   = {12'h003};
-    //     bins frm    = {12'h002};
-    //     bins fflags = {12'h001};
-    // }
     instrs: coverpoint ins.current.insn {
         wildcard bins fsw          = {32'b????????????_?????_010_?????_0100111};
         wildcard bins flw          = {32'b????????????_?????_010_?????_0000111};
@@ -90,6 +82,7 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
         wildcard bins fsqrt_dyn    = {32'b01011_??_00000_?????_111_?????_1010011};
         wildcard bins fround_dyn   = {32'b01000_??_00100_?????_111_?????_1010011};
     }
+
     // main coverpoints
     cp_mstatus_fs_illegal_instr: cross instrs, mstatus_FS_zero;
     // cp_mstatus_fs_csr_write:  redundant, covered by cp_mstatus_fs_illegal_instr
@@ -97,5 +90,8 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
 endgroup
 
 function void exceptionsf_sample(int hart, int issue, ins_t ins);
+    $display("Mstatus FS: %b, frmIllegal: %b, op: %b, fmrBits: %b, imm: %b", ins.current.csr[12'h300][14:13], ins.current.csr[12'h003][7:5],  ins.current.insn[6:0], ins.current.insn[14:12], ins.current.insn[31:27]);
     ExceptionsF_exceptions_cg.sample(ins);
+
+    
 endfunction
--- a/fcov/priv/ExceptionsF_coverage.svh
+++ b/fcov/priv/ExceptionsF_coverage.svh
@@ -90,7 +90,7 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
 endgroup
 
 function void exceptionsf_sample(int hart, int issue, ins_t ins);
-    $display("Mstatus FS: %b, frmIllegal: %b, op: %b, fmrBits: %b, imm: %b", ins.current.csr[12'h300][14:13], ins.current.csr[12'h003][7:5],  ins.current.insn[6:0], ins.current.insn[14:12], ins.current.insn[31:27]);
+    // $display("Mstatus FS: %b, frmIllegal: %b, op: %b, fmrBits: %b, imm: %b", ins.current.csr[12'h300][14:13], ins.current.csr[12'h003][7:5],  ins.current.insn[6:0], ins.current.insn[14:12], ins.current.insn[31:27]);
     ExceptionsF_exceptions_cg.sample(ins);
 
     
--- a/tests/lockstep/priv/ExceptionsF.S
+++ b/tests/lockstep/priv/ExceptionsF.S
@@ -10,10 +10,44 @@
 
 #include "WALLY-init-lib.h"
 
+//Function that executes the floating point arithmetic and conversions instructions
+fp_instr:
+    # --- Floating‑Point Arithmetic ---
+    # fadd: f2 = f0 + f1
+    fadd.s f2, f0, f1        
+    # fsub: f3 = f0 - f1
+    fsub.s f3, f0, f1        
+    # fmul: f4 = f0 * f1
+    fmul.s f4, f0, f1        
+    # fdiv: f5 = f0 / f1
+    fdiv.s f5, f0, f1        
+
+    # --- Floating‑Point Conversions ---
+    # Convert float to integer: x2 = int(f0).
+    # (Using fcvt.w.s here; some assemblers support fcvt.x.f as an alias.)
+    fcvt.w.s t2, f0          
+    # Convert integer back to float: f6 = float(x2).
+    fcvt.s.w f6, t2          
+    # Convert float to float (no‑op conversion): f7 = f0 converted (using a pseudo instruction).
+    fcvt.s.d f7, f0          
+
+    # --- Fused Multiply-Add and Other Ops ---
+    # fmadd: f8 = (f0 * f1) + f2
+    fmadd.s f8, f0, f1, f2    
+    # fsqrt: f9 = sqrt(f0)
+    fsqrt.s f9, f0          
+    # fsgnj: f10 gets f0’s magnitude with f1’s sign.
+    fsgnj.s f10, f0, f1 
+    # fround: round f0 and place result in f14 (pseudo‑instruction; actual support may vary).
+    fround.s f14, f0      
+
+    //return
+    ret
+
 main:
-/////////////////////////////////
-//cp_mstatus_fs_illegal_instr
-/////////////////////////////////
+///////////////////////////////////////////////////////////
+//cp_mstatus_fs_illegal_instr AND cp_mstatus_fs_csr_write
+///////////////////////////////////////////////////////////
     # set mstatus.FS to 00 to disable fp
     li t0, 0
     csrs mstatus, t0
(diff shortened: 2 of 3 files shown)


tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+///////////////////////////////////////////
+
+#include "WALLY-init-lib.h"
+
+//Function that executes the floating point arithmetic and conversions instructions
+fp_instr:
+    # --- Floating‑Point Arithmetic ---
+    # fadd: f2 = f0 + f1
+    fadd.s f2, f0, f1
```
jordancarlin, line 17: The comments that are just repeating the instruction itself seem unnecessary.

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+    fcvt.s.d f7, f0          
+
+    # --- Fused Multiply-Add and Other Ops ---
+    # fmadd: f8 = (f0 * f1) + f2
+    fmadd.s f8, f0, f1, f2    
+    # fsqrt: f9 = sqrt(f0)
+    fsqrt.s f9, f0          
+    # fsgnj: f10 gets f0’s magnitude with f1’s sign.
+    fsgnj.s f10, f0, f1 
+    # fround: round f0 and place result in f14 (pseudo‑instruction; actual support may vary).
```
jordancarlin, line 40: fround is not a pseudoinstruction. It is a regular instruction but it is part of the Zfa extension. Not sure that we should be including it in this test that should only depend on the F extension.

fcov/priv/ExceptionsF_coverage.svh
```diff
@@ -25,25 +25,17 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
...
 
     // building blocks for the main coverpoints
-    mstatus_FS_zero: coverpoint ins.prev.csr[12'h300][14:13] {
+    mstatus_FS_zero: coverpoint ins.current.csr[12'h300][14:13] {
         bins disabled = {2'b00};
     }
-    mstatus_FS_nonzero: coverpoint ins.prev.csr[12'h300][14:13] {
-        bins enabled = {!2'b00};
+    mstatus_FS_nonzero: coverpoint ins.current.csr[12'h300][14:13] {
+        bins enabled = {2'b11};
```
jordancarlin, line 32: Why is this switching from anything other than off to specifically 0b11? Anything other than 0 indicates the F extension is enabled.

fcov/priv/ExceptionsF_coverage.svh
```diff
@@ -25,25 +25,17 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
     option.per_instance = 0; 
 
     // building blocks for the main coverpoints
-    mstatus_FS_zero: coverpoint ins.prev.csr[12'h300][14:13] {
+    mstatus_FS_zero: coverpoint ins.current.csr[12'h300][14:13] {
```
jordancarlin, line 28: I'm not sure I understand why all of these are switching from prev to current. Don't we want to know the CSR value when the instruction starts, not when it finishes?

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+    //return
+    ret
+
+main:
+///////////////////////////////////////////////////////////
+//cp_mstatus_fs_illegal_instr AND cp_mstatus_fs_csr_write
+///////////////////////////////////////////////////////////
+    # set mstatus.FS to 00 to disable fp
+    li t0, 0
+    csrs mstatus, t0
```
jordancarlin, line 52: This doesn't seem like it actually does anything? csrs sets any bits that are ones in t0 to ones in mstatus. It doesn't change any bits that are already one to zero. I think you need a csrc to the bits of mstatus that FS is in.

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+
+
+    # --- Floating-Point Memory Operations ---   
+    # Store f0 into memory at scratch.
+    fsw    f0, 0(a4)          # fsw: store word from f0 into address a4+0
+
+    # Load a floating‑point value from memory into f1.
+    flw    f1, 0(a4)          # flw: load word from address x1+0 into f1
+
+    # Execute floating point extructions from function
```
jordancarlin, line 89: typo

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+    fmin.s f12, f0, f1      
+
+    # --- Additional Floating‑Point Instructions ---
+    # fli: load immediate float 2.5 into f13.
+    fli.s    f13, 2.5              
+
+    # --- CSR Operations on Floating‑Point Status Registers ---
+    # Use csrrw to read and write the FCSR.
+    csrrw  x7, fcsr, x0     # csrrw_fcsr: exchange x7 with fcsr (writing x0)
+    # Assume pseudo‑CSRs “frm” for rounding mode and “fflags” for exception flags.
```
jordancarlin, line 114: These are actual CSRs, not pseudo‑CSRs. They just shadow fcsr.

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+
+    //Initialize scratch memory
+    li     t1, 0xDEADBEEF
+    sw     t1, 0(a4)
+    sw     t1, 4(a4)
+    sw     t1, 8(a4)
+    sw     t1, 12(a4)
+
+    //initialize floating point registers used
+    fli.s    f0,  1.0            # fli: load immediate float (1.0) into f0
```
jordancarlin, line 65: Not sure that it makes sense to initialize all of the floating point registers to the same value. That makes it tricky to check if instructions are succeeding/failing as expected because for many instructions the result will be the same either way.

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+    fli.s    f1,  1.0               
+    fli.s    f2,  1.0
```
jordancarlin, line 67: The fli instruction is part of the Zfa extension, not the F extension. You can't rely on it being present. You'll need to use instructions that are present in the F extension, so either initialize these registers with values from memory or load immediates into integer registers and then move/convert those values into the floating point registers.

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,167 @@
...
+    la      a4, scratch
+
+    //Initialize scratch memory
+    li     t1, 0xDEADBEEF
+    sw     t1, 0(a4)
+    sw     t1, 4(a4)
+    sw     t1, 8(a4)
+    sw     t1, 12(a4)
+
+    //initialize floating point registers used
```
jordancarlin, line 64: You are disabling floating point before initializing these. Won't that make these instructions all trap? I think you want to enable floating point, initialize the registers, and then disable floating point to test the instructions that you want to trap.

jordancarlin (changes requested):

C0d3ing:
Thank you for the Feedback, made the following changes

ExceptionsF_coverage.svh:

- Reverted ins.current to ins.prev
- Corrected mstatus_FS_nonzero since {!2b'00} was not working as expected
- Removed fli and fround instructions from coverpoints since these are part of the Zfa extension

ExceptionsF.S:

- Initialized floating point registers with fcvt.s.w to different values instead of fli.s
- Updated mstatus.FS mask to enable and clear properly
- Enabled mstatus.FS for initialization then disabled after
- Removed redundant comments
- Removed fli and fround instructions (part of Zfa instruction set)

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,176 @@
...
+    #ifdef __riscv_xlen
+    #if __riscv_xlen == 32
+        # Extract the lower 32 bits (bits 31:0) into t1.
+        fmv.x.w t1, f0
+        
+        # Extract the upper 32 bits (bits 63:32) into t2.
+        fmvh.x.d t2, f0
+
+        #Reassemble the double from t1 and t2 back into a floating-point register (f1).
+        fmvp.d.x f1, t1, t2
```
jordancarlin, line 146: `fmvh.x.d` and `fmvp.d.x` depend on both the Zfa and D extensions. Do we want to test these?

There is an argument to be made that since this is a test that generates exceptions anyways, it doesn't hurt to include these instructions because they will just trap on an illegal instruction exception. If we go that route, then `fround` can be retained as well. Not sure this is necessary though. @davidharrishmc any thoughts?

C0d3ing:
ExceptionsF_coverage.svh:

- Included Zfa dependencies in cover groups

ExceptionsF.S:

- Included Zfa assembly test code

jordancarlin (approved):
Looks good to me!

tests/lockstep/priv/ExceptionsF.S
```diff
@@ -0,0 +1,180 @@
...
+    sw     t1, 0(a4)
+    sw     t1, 4(a4)
+    sw     t1, 8(a4)
+    sw     t1, 12(a4)
+
+    //initialize floating point registers used 
+    li t0, 1
+    fcvt.s.w f0, t0
+
+    addi t0,t0,1
```
davidharrishmc, line 58: These addi statements would be clearer as li t0, 2; li t0, 3, etc. so it's obvious what is in the f registers.

C0d3ing:
ExceptionsF:
- changed the addi fp initializations to li instructions

C0d3ing:
ExceptionsF_coverage.svh:

- Included new coverpoint cp_mstatus_fs_legal

ExceptionsF.S:

- Moved the floating point instructions into a function call. This reduced similar code between cp_mstatus_fs_illegal_instr, cp_mstatus_fs_csr_write, and cp_mstatus_fs_legal

C0d3ing:
ExceptionsF_coverage.svh:

- Created a new covergroup and cross product to match testplans 
- removed add from the instr coverpoint since it was not a floating instruction

ExceptionsF.S:

- Modified both function to accept desired Mstatus.FS testing mode.  To bypass the automatic FPU updates on Mstatus.FS, Mstatus.FS is set before each test function is called.

davidharrishmc:
Now that I'm not on a bus, I gave this a closer read for style.  Here are a bunch of style improvements:

Space between // and comment
Be consistent of // vs # for comments – pick // 

16: move this li into  fp_instructions; it is hard to find the value in s1 when it is a global variable.
I’d mentioned having a legal frm as well to make sure the instructions don’t just always fault.  in ExceptionsF_coverage frm_illegal, add a bin with, say, 0.
Lines 57-67 say illegal instructions, but the instructions themselves are legal, and the call is with legal valus of FS.
Why not make the argument to fp_instructions be in a0 to follow standard calling conventions?
Comments such as at 58 “load a1 to write 1” say what the code is doing, which is obvious from li a1, 1.    It would be better to say “Test floating-point instructions with mstatus.FS = 01”
Comments such as 164 “function return” don’t add anything to the ret statement.
86: “Set up correct Mstatus.fs mask” -> “Shift a1 (becomes a0) into position to write mstatus.FS”  (note capitalization consistent with priv spec and book, be consistent through your file)
Worth doing a close proofread on the code, and review of your comments for style and clarity.
Why does badfrm only test fp_arith_instr, while the others test fp_instrs?  Can you use fp_instrs for both and eliminate fp_arith_instr (move into fp_instrs at 261?  Then you could get rid of dyn_instrs in the coverage file and change cp_badfrm to cross with instrs, showing that the rounding mode doesn’t affect certain instructions.
“This ensures the automatic FPU updates to Mstatus.FS are bypassed” -> unclear what you mean by bypassed.  Are you trying to say the code sets mstatus.FS to the intended value before each test because the previous test may have marked it dirty?

C0d3ing:
General:

- Consistent use of // instead of # 
- Removed spaces after //

ExceptionsF_coverage.svh:

- Removed the dyn_instrs coverpoint
- Updated the coverpoint badfrm to be crossed with instrs
- added frm_legal
- cp_mstatus_fs_legal is now crossed with frm_legal


ExceptionsF.S:

- Moved global variable s1 into fp_instructions
- mstatus.FS value argument was changed from a0 to a1 to follow standard calling conventions
- Added new argument to fp_instructions, a1, to set desired frm value.
- Merged fp_arith_instr into fp_instructions
- Updated comments: addressed concerns for lines (57-56, 58, 164, 86), generally improved readability, removed comment redundancy

C0d3ing:
ExceptionsF_coverage.svh:

- Updated bin name for the coverpoint frm_legal
- Coverpoint now allows for any legal rounding mode (0 through 4)

davidharrishmc:
still missing space between // and comments
Lines 15-66 are clearly organized and easy to follow.  
73: the helper function is fp_instructions.  Just call it that in the comments.
77: comments are too long and wrap in an ugly way.  Break lines.
The first sentence says all the floating point instructions in the F extension.  But the program just tests selected instructions.  You haven’t explained that this is intended to test the instructions with different mstatus.FS and with legal and illegal rounding modes.
Then “(Instructions from Zfa extension are labeled… “ These are not from the F extension, so it would be clearer to say “Also test selected Zfa instructions, which will trap if Zfa is not supported.”
The description of a0 and a1 is ackward. You aren’t clear these are input arguments.  They are described with a verb “write” but are nouns (“value to write”) The intro sentence doesn’t explain why you are writing these two values.

155: Why is it important to call out “After initialization”?  Wouldn’t this work at any time?
Please proofread.  “Set up a pointers” “Wrtite the value” “. program”
163, 169, etc.  If you get rid of these blank lines, it will be clearer that the floating-point instruction is associated with each mstatus.
320- csrrs with x0 and csrrc with x0 don’t set or clear anything, and could be treated as just csrr by some impelementations.  It would be more robust to have a nonzero value, such as 1, to set and clear.
It would be clearer to group together all the Zfa extension tests rather than scatter them.

C0d3ing:
ExceptionsF.S:

- Addressed comments for lines 73, 155
- Updated fp_instructions header
- removed blank lines after each mstatus.FS set (lines 163, 169, ...)
- All cssrs and cssrc test instruction (320 -) now have non zero write values

PR #537, opened 2025-03-22
Testbench for Sail coverage

jordancarlin:
Initial version of a testbench for coverage collection using a log file instead of an RTL DUT. The testbench expects the log to be in a specific format. Python script to create this format is still a work in progress and will be part of a follow up PR (I have a very hacky version working for testing but it needs significant cleanup). A do file is also in the works for running this simulation along with another python script that will automate the process of running a set of elfs on Sail, converting the log files, and running the coverage simulation. I wanted to get some early feedback on the testbench structure before proceeding further with all of the supporting scripts.

Using the hacky scripts with this testbench, I was able to achieve near full coverage of the unprivileged coverpoints (and many privileged coverpoints).


jordancarlin:
Initial version of .do file added.

PR #538, opened 2025-03-24
SsstrictM and SsstrictS covergroups

coreyqh:


jordancarlin (commented):
Looks like this is testing branches and jumps right now, which are likely to cause problems. We want to test any illegal encodings within the branch/jump major opcode spaces but not legal jumps because they'll take us to a different part of the program.

Also, this PR should pull the coverpoints out of the existing Exceptions covergroups. Those versions might be a good reference as you flush this version out.

davidharrishmc:
I was thinking that SstrictM would just move over the existing coverpoints of ZicsrM_mcsr, ExcceptionsM_instr, and ExceptionsZc_instr.  Have I forgotten something?

davidharrishmc:
I don't remember them missing anything, but if you have any notes, it would be worth checking.

davidharrishmc:
Can you call out what changes were made besides copying the coverpoints to the new file?  The only one I noticed on a quick inspection is changing priv mode m to msu for the CSR tests.

Should we be clear about whether the instruction coverage is intended for machine or supervisor mode or both?   Should there be a SsstrictM and SsstrictS, or should this just run in supervisor mode?  Should the CSR tests only run in the privilege mode of the suite?

Do we need to remove the corresponding tests from ZicsrS and ZicsrU and from their testplans?

coreyqh:
Reworked to remove the mcsr covergroup from SsstrictS and to avoid repeated code using `include files for the instruction coverage.

PR #541, opened 2025-03-27
ZicsrF (100% Coverage)

C0d3ing:
Since ZicsrF is only testing 3 CSRs I did not create a header-generating function in the csrtests.py file. Please let me know what you prefer @davidharrishmc.  

ZicsrF_coverage.svh:

- Removed add from instrs
- Corrected funct3 values for csrr instructions
- changed walking ones bin to always be 32 bits. Since it seems like we are only testing single precision instructions here and the D extension is not guaranteed.
 
ZicsrF_coverage.svh:

- Initial assembly test
- Borrowed fp_instructions from ExceptionsF. Set a1 input as 0 to ensure frm is always legal.


davidharrishmc:
For walking ones, restore to the XLEN code you used to have.  We should test writing all bits. In the three walk routines, instead of having a shift by t0, load t5 with 1 initially, and shift it left one on each step until you terminate when t0 becomes 0 (the 1 shifts off the end).

466: “set” has ambiguous meaning.  One meaning is to make bits 1.  So it would be clearer to say either clear mstatus.FS or mstatus.FS=00.  The later choice works for all of the test cases, with the different FS values.

468: it would be nice to add comments like inf – inf sets invalid flag.  Same for each instruction that should set a flag

PR #544, opened 2025-03-31
Assembler to make illegal instructions easy to write coverage tests for

jacassidy:


jordancarlin:
I don't want to waste all of the work you've done on this, but take a look at the `.insn` directive that can go directly into an assembly file: https://sourceware.org/binutils/docs/as/RISC_002dV_002dFormats.html

Not sure if that is sufficient, but if it is then it would be better to avoid maintaining another tool.

jordancarlin:
@jacassidy Based on the conversation we had, I think this can be closed?

PR #548, opened 2025-04-01
Move include files to priv directory

coreyqh:
Reopened due to ECA issues faced yesterday.


coreyqh:
Also removed ucsr covergroup from ZicsrU

PR #550, opened 2025-04-06
Changed and new coverpoint templates according to new plan

georgiatai:
Some changes made according to the new testing plan and fixes:
- Deleted the coverpoints for comparing eqval
- Some typos fixed and changed all csrs to be sampled before rather than after
- Added more concrete templates (those that do not require python script), e.g. vlmul int values, and vstart of 0


jordancarlin:
@georgiatai It looks like all of these changes are included in #560 now. Can you confirm, and if so, close this PR?

PR #554, opened 2025-04-09
Exceptions zalrsc

C0d3ing:
fcov/coverage/RISCV_coverage_config.svh:

- Included in coverage

fcov/priv/ExceptionsZalrsc_coverage.svh:

- sc.x writes a non-zero value when it fails to store to a reserved address. It also fails when the address is misaligned or on an instruction access fault. Thus the original cross product `cp_store_address_misaligned: cross sc, adr_LSBs, rd_boolean;` was impossible to hit (can't have rd_boolean = 1 and misaligned instructions). This is also true for `cp_store_access_fault`. So I split the cover points into cases where the LSBs are legal offsets in RV32/64. Then I crossed it with the expected rd value.

tests/lockstep/priv/ExceptionsZalrsc.S:

- Initial assembly test


davidharrishmc:
ExceptionsZalrsc_coverage.svh
I updated the two sc coverpoints in the spreadsheet to cross with rd having an initial value > 1, to check that faulting sc should not change the value of rd.
I added two more coverpoints in the spreadsheet for simultaneous misaligned and access faults to check priority of the faults. Can you add these to the coverage.svh? We can implement them with an access to ACCESS_FAULT_ADDRESS+1.
ExceptionsZalrsc.S:  
16, 37: these headers should also mention the store coverpoints which are tested here too.
21-33: can you keep the loop counter in a0 and not use a2 and not need the mv a0, a2?
48: function name is outdated

C0d3ing:
ExceptionsZalrsc_coverage:
Testing that Ir address robustly matches the sc address doesn't seem easy. I could check current rs1 and prev rs1 to see if the values match but if a trap occurs then this wouldn't work (happens when loading from misaligned value). We could also just write the test so that the address matches and is easy to see by inspection.  I would need to comment on the cover group and test file about this.

I also added a check to see if the rd register is an expected value after the sc.x instruction. If the address offset is illegal then rd prev and current should not change. But when a legal offset is used then rd current should become 0. 

ExceptionsZalrsc.S:

- updated headers, function name, and loop counter register

C0d3ing:
ExceptionsZalrsc_coverage:

- Added `cp_load_misaligned_priority:               cross lr, illegal_address_misaligned;` to collect coverage on when lr.x tries to load from an address of (ACCESSFAULT + 1)

PR #557, opened 2025-04-10
Align coverpoint for I instructions

ahlyssa:
Screenshot 2025-04-10 at 3 54 57 PM

- edits to testgen to handle loaditypes and stypes correctly (removed unnecessary addi)

- all align coverpoints are at 100%


jordancarlin:
Slightly concerned about the changes to testgen beyond adding the new coverpoint because of likely conflicts with the signature generation work Marina is doing, but it does seem like a nice cleanup.

davidharrishmc:
@ahlyssa did you test your change on the roll regression-wally --fcov?  I think it will break the regular load and store offset tests because those will now be misaligned.

PR #560, opened 2025-04-12
Vector Coverage Running

georgiatai:


bin/covergroupgen.py
```diff
@@ -87,6 +90,17 @@ def customizeTemplate(covergroupTemplates, name, arch, instr):
...
 
+def customizeEFFEWTemplate(covergroupTemplates, name, effew):
+    if (name in covergroupTemplates):
+        template = covergroupTemplates[name]
+    else:
+        if (not (name in missingTemplates)):
+            print("No template found for " + name)
+            missingTemplates.append(name)
+        return ""
+    template = template.replace("EFFEW", effew)
```
jordancarlin, line 101: I wonder if we'd be better off just adding this as an additional replacement in the existing `customizeTemplate` function. Are you expecting to add a bunch of other things to it? Because right now it is just a single new line and probably doesn't warrant duplicating the rest of the function.

fcov/coverage/RISCV_coverage_common.svh
```diff
@@ -366,6 +366,45 @@ function bit get_vm(string s);
   endcase
 endfunction
 
+//Vector vsetvli paramaters
+// Flipped: string -> int (bits), using case statements
+
+function bit [2:0] get_vtype_eSEW_val(string str);
```
jordancarlin, line 372: For consistency, it might be better to go with `get_vtype_eSEW` (and the same for the other added functions).

georgiatai: In dissassemble_helpers there are similar functions (get_vtype_eSEW_name) that does the opposite of the Val function here. We added _val mainly to distinguish the two similar functions. Do you have a different proposed name or way to do this?

jordancarlin: We'll leave it for now and can tweak later.

PR #564, opened 2025-04-15
Vector Preliminary Exceptions coverpoints

jacassidy:
Not sure if it makes sense to merge at this time since its largely incomplete but these are successfully running vector exception coverpoints


davidharrishmc:
This has 300 files, most of which are not your work. Close this out and
regenerate the PR with just the files that you want to merge.

PR #565, opened 2025-04-15
First Draft Vector Priv Coverpoints

jacassidy:
running with regression but not anywhere near finished


jacassidy:
I know I uploaded two other files by accident ill fix later

fcov/priv/ExceptionsV_coverage.svh
```diff
@@ -0,0 +1,525 @@
...
+// either express or implied. See the License for the specific language governing permissions 
+// and limitations under the License.
+////////////////////////////////////////////////////////////////////////////////////////////////
+
+`define COVER_EXCEPTIONSV
+
+covergroup ExceptionsV_exceptions_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+
+    vtype_prev_vill_clear: coverpoint ins.prev.csr[12'hC21][`XLEN-1] {
```
jordancarlin, line 28: Especially when it is a CSR field that moves with different xlens I recommend using the `get_csr_val` function instead. It lets you specify a CSR name and field.

jordancarlin (commented):
Make sure to add this to the config file with the `include.

jacassidy:
[log.txt](https://github.com/user-attachments/files/19762251/log.txt)
Regression ran successfully but the log file has errors that I'm not sure why pop up as the covergroups for vector shouldnt have been sampled

davidharrishmc:
What does this need before merge?  @jordancarlin have you finished review?

jordancarlin (commented):
Sorry for the delayed review.

Fine to merge this if we want for now, but it seems like a lot of these coverpoints should probably be recategorized. For example, `cp_vcsrrwc` probably belongs in a `ZicsrV` covergroup, as does `cr_vset_i_vl_i_lmul_sew`.

We also shouldn't be testing reserved values outside of the Ssstrict extension, since any behavior is legal (ie the `csrops_reserved_vcsrs` should be removed).

PR #567, opened 2025-04-16
Align coverpoints for I instructions 

ahlyssa:
- this pr contains the coverpoints as well as templates for the align holes


davidharrishmc:
Looks like sw needs a word entry.

Issue #568, opened 2025-04-17
Zicsr* tests and coverpoints haven't been moved properly after repartitioning with Ssstrict


davidharrishmc:
Zicsr_coverage lost walking 1s.  Need to put them back for all three files for registers at that privilege level.
ZicsrU.S has ZicsrU-CSR-Tests, which aren’t required anymore because they moved to Ssstrict, but no walk tests.
ZicsrM.S should be able to drop ZicsrM-CSR-Tests, which moved to Ssstrict.
ZicsrS.S should be able to drop ZicsrS-CSR-Tests.h, which moved to Ssstrict
Check testplans match this.

ZicsrM cp_mret and cp_sret depend on mpp and spp having values that might not exist.  Need to move some of this coverage into ZicsrS/U executed in machine mode

davidharrishmc:
Partially fixed with 575.  cp_mret/sret still pending

davidharrishmc:
Done

PR #575, opened 2025-04-22
Moved walking 1s back to Zicsr{M,S,U} covergroups

coreyqh:
* See #568 
* Zicsr* was missing writes to M,S,U CSRs in the corresponding mode. For M and S these were incorrectly moved to Ssstrict, and for U they were dropped


coreyqh:
Moved {m/s}ret coverpoints from M to S, and replaced them with a simpler version that ignored mpp and spp. 

Added tests for Zicsr{m/s/u} walking ones coverpoints commited earlier

Dropped Zicsr*-CSR-Tests.h from Zicsr* as these were moved to Ssstrict

davidharrishmc:
In ZicsrM_coverage, I think we should keep old_status_mpp but only have the M_mode bin.  

iin ZicsrM_coverage, I don’t think we can cross with old_mstatus_tsr because it is read-only 0 when S-mode is not supported.

Do the tests hit all these new coverpoints?

coreyqh:
Made suggested cross product updates

I have full coverage of Zicsr*

PR #576, opened 2025-04-22
fixes to interruptsS coverage, removing delegation of machine interrupts

HamzaYusufJamal:
Our coverpoints tested for mideleg bits for M-mode interrupts being high, but we can't test this on Wally. The coverpoints are now agnostic to whether M-mode interrupts can be delegated, but only check cases for S-mode interrupts.


davidharrishmc:
There's a merge conflict to fix, and then this can be merged.

PR #581, opened 2025-04-29
Align Coverpoint Hole for I instructions

ahlyssa:
- initially was not working due to the extra space in the testplans after cp_align
- all 3 instructions are a 100%


davidharrishmc:
It looks like lb, lbu, and sb are identical.  Why do you need all three?  Can you use just one for all three places to reduce the number of files?  Also, the files don’t match the header, such as align_word.S, and these headers are inconsistent.
Same with lh, lhu, sh, and with lw, lwu, sw.
Why is lw different than the others with the xlen and sd business?  Can’t you use the sw to initialize memory, and skip the sd?
What is the point of the lw in the following code?  Can you drop it without loss of coverage?
#ifdef __riscv_xlen
#if __riscv_xlen == 64
lwu t3, 4(t0)
#else
lw t3, 4(t0)
#endif

PR #583, opened 2025-05-01
ExceptionsZicboU/S Coverage

C0d3ing:
Initial Draft for ZicboU/S test files.

### DIFF
--- a/fcov/coverage/RISCV_coverage_config.svh
+++ b/fcov/coverage/RISCV_coverage_config.svh
@@ -239,3 +239,9 @@
 `ifdef EXCEPTIONSV_COVERAGE
   `include "ExceptionsV_coverage.svh"
 `endif
+`ifdef EXCEPTIONSZICBOU_COVERAGE
+  `include "ExceptionsZicboU_coverage.svh"
+`endif
+`ifdef EXCEPTIONSZICBOS_COVERAGE
+  `include "ExceptionsZicboS_coverage.svh"
+`endif
--- a/fcov/priv/ExceptionsZicboS_coverage.svh
+++ b/fcov/priv/ExceptionsZicboS_coverage.svh
@@ -52,9 +52,9 @@ covergroup ExceptionsZicboS_exceptions_cg with function sample(ins_t ins);
     }
 
     // main coverpoints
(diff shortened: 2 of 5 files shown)


C0d3ing:
tests/priv/ExceptionsZicbo{S/U}.S:

- Initial Assembly test

fcov/coverage/RISCV_coverage_config.svh:

- Included in coverage collection

fcov/priv/ExceptionsZicboS_coverage.svh:

- Wildcard bins update
- privmode update

fcov/priv/ExceptionsZicboU_coverage.svh:

- Wildcard bins update

davidharrishmc:
Comments apply to both files but are listed for ExceptionsZicboS
Why are the ecalls necessary in lines 17-74?  Seems like the functions already do this, so you could drop this?
“load into machine mode” -> “change into machine mode”
259: replace with “larger than any reasonable cache line to prevent”
129-139: I don’t think you need any of this because you will explicitly set the mode and xenvcvg registers before testing again anyway.
123, 127: code says i/j < 4, but line 93 says 2.  Needs to be consistent; I think you want 4?  I don’t understand how you are getting 100% coverage with the wrong value here; is there a bug in the coverpoint?  It’s important to get to the root cause here.  
182-186, 238-242: now I think the comments are wrong
all programs: saving the priv mode should be at the start of the function outside the loops.  Or you could just pass the argument in a1 and eliminate the save code (but correct the comments).
The la t2, DataBlock might as well also be moved to the top of the functions so it is only executed once.
It would be more conventional to use s registers for the loop variables.

C0d3ing:
For both ZicboU/S:

- Removed the redundant ecalls when calling the function
- Updated comment “load into machine mode” -> “change into machine mode”
- The loop limit mismatching the comments was a mistake forgot to update the comments as I copy pasted the sections. Now they should be correct.
- Desired priv mode for testing is now passed in argument a1 to remove an extra mv instruction
- la t4, DataBlock has been moved outside of the loop 
- loop counters use the s registers now

davidharrishmc:
Why do you need a load into machine mode at both the start and when returning?  I think you could drop one.

C0d3ing:
I removed the ecall to load into machine mode privilege at the end of each function in both ZicboU/S and updated each function's header.

PR #585, opened 2025-05-01
PMPM Initial Coverpoints

sudo-apt-abdullah:
This PR is made so that the coverpoints written so far can be reviewed before proceeding further and we can get the feedback if something can be done in a better way.


davidharrishmc:
This looks like a very promising approach.

Lines 21-27 need a lot more explanation.  There are probably some assumptions about alignment in these numbers.
I don’t think line 27 is right.
Doesn’t look like Mcause is needed anymore
standard_region relates to region 0, while locked_region and subsequent ones seem to relate to region 1.
I think the pmp_hit field of coverpoints like X0 should have 0 instead of ? in the less significant bits, so that if we match both region 4 and 7, only 4 shows up as hit.
For cp_cfg_L*_pmp*, the coverpoints don’t match up with cp_cfg_L_modify.  The coverpoint is trying to check the effect with pmp*_write_ignore/detect.  But all we really need to do is perform the operation, and the lockstep or self-check will check the effect.  So I think the test could involve a write to pmpcfg_i, and no check of what happens to pmpcfg.  It doesn’t look like the coverpoint is checking pmpcfg_i.XWR or pmpaddr_i.

PR #587, opened 2025-05-02
PMPM Initial Coverpoints

sudo-apt-abdullah:
Thanks for the review and feedback as we need a lot of. I tried to refined the covergroup and fixed the pmp_hit logic and the coverpoint cp_cfg_L_modify.


davidharrishmc:
Does STANDARD_REGION work for G = 0?  What does G-1 produce?
It seems like the appropriate grain for NAPOT (not NA4) is max(G,1).
Add comments about why lock_checking and pmp_region related to [1] rather than [0]?  I think it is because you share them for TOR where you need region 0 too.
Some of the indentation is funky now
You may need to us ins.prev instead of ins.cur to check the value of a CSR at the start of the instruction rather than the end.  That will matter for things like cp_cfg_L_modify, which try to change the CSR.
cp_cfg_L_modify should be crossed with the values in pmpcfg[1] and pmpaddr[1], and the value in rs1 being written or cleared, so the entire feature in the spreadsheet is exercised.
I don’t think we need an ignore_bins in cp_cfg_L_modify_TOR, in that it is good to test both locked and unlocked. I’ve amended the spreadsheet to list both L = 0 and 1.
We now have a pre-commit check that complains about trailing spaces on lines.  Please view details on the failed check and clean that up.
Do the tests hit these coverpoints?  If they do, let’s merge this PR.

davidharrishmc:
G will eventually need work because Wally uses G=0 and we should be able to fully test it.  However, the standard region will be a minimum of 8 bytes, and we only need to test G=0 for certain nonstandard regions

Indenting within pmp_region, write_*_csr is still inconsistent

write_prev_pmp_csr is a confusing name.  I thought it referred to ins.prev for a while.  How about write_lower_pmp_csr?

Does the pmp_sample function need ins.prev?  Is it trying to sample the value before or after the operation?

I think you want a bigger cross to check all the conditions:

write_cfg_L_modify: cross lock_checking, pmp_region, write_pmp_csr, pmpcfg_xwr, pmpaddr, rs1vals;

pmpcfg_xwr: coverpoint pmpcfg[0][6:0] {
	bins NAPOT_XWR = {7’b0011111};
}
Note***: write_pmp_csr works in combination with rs1vals
write_pmp_csr: coverpoint ins.prev.insn {
      wildcard bins write_pmpaddr  = {32’b001110100001_?????_010_?????_1110011}; // Try to write pmpaddr[1]
        wildcard bins write_pmpcfg   = {32’b001110110001_???_010_?????_1110011}; // Try to write pmpcfg[1]
	}
pmpaddr: coverpoint ins.prev.csr[12’h3B0] {
	bins four = {4};
}

rs1vals: coverpoint ins.current.rs1_val {
	bins zero: {0};
	bins lockoff: {15’b0011111_00000000}; // region 1 lock off
}

similar for cp_cfg_L_modify_TOR

sudo-apt-abdullah:
Seems good, I've updated the pmpaddr value to 'd256/'h100.

The write_pmp_csr clear_lock_bit bin doesn’t have enough information to check that the test is really trying to clear the lock bit. *** drop?
Yes, you're right regarding clear_lock_bit bin, it depends on rs1_val which we don't for the time being, but 
wildcard bins write_pmpcfg   = {32'b001110110001_00000_010_?????_1110011};
this will try clear lock as well. So, it's better to drop that. (Already done)

Sure, I'll try to put these in a separate file to make it decent. I'll rename it as well to match with the name in the testplan. And to split the coverpoint, I'll look into that as well.

Yes, merging it will be good because Hamza also need to make a PR, but waiting for this PR to be merged. :)

PR #588, opened 2025-05-07
Checked extension M and updated testgen.py to handle when we have more than one extension

Marina-Bellido:
Checked extension M (100% coverage for rv32 and rv64) and updated testgen.py to handle when we have more than one extension


templates/testgen/testgen_footer.S
```diff
@@ -17,8 +17,6 @@ RVTEST_DATA_BEGIN
 rvtest_data:
 .word 0xbabecafe
 .word 0xbabecafe
-.word 0xbabecafe
```
jordancarlin, line 20: Are you intentionally removing these lines? And if so, why? This seems unrelated to supporting M.

bin/testgen.py
```diff
@@ -20,20 +20,36 @@
...
+    ext_regex = ".*I.*" + "".join([f"{ext}.*" for ext in ext_parts])  # ".*I.*Zfa.*D.*"
+    # Final TestCase line
+    test_case_line = f"//check ISA:=regex(.*{xlen}.*);check ISA:=regex({ext_regex});def TEST_CASE_1=True;"
+    if extension == "I":
+      # Replace placeholders with the actual values
+      template = template.replace("sigupd_count", str(sigupd_count))
+      template = template.replace("ISAEXT", f"RV{xlen}{extension}")
+      template = template.replace("TestCase", f"//check ISA:=regex(.*{xlen}.*);check ISA:=regex(.*{extension}.*);def TEST_CASE_1=True;") # , str(instruction)
+      template = template.replace("Instruction", test)  #missing the 0 in front check meeting
+    else:
```
jordancarlin, line 46: There is a lot of duplication between the if and the else cases here. Could you just remove the duplicate I from the list when the extension is I and then not need to have all of the rest of this in a conditional?

templates/testgen/testgen_footer.S
Marina-Bellido: yeah I am, prof Harris said it was not necessary to have some many .word 0xbabecafe if they were not being used

bin/testgen.py
Marina-Bellido: Just pushed a simplified version with no conditionals.

bin/testgen.py
```diff
@@ -1840,8 +1847,6 @@ def write_tests(coverpoints, test, xlen):
       pass # Zalrsc coverpoints handled custom
     elif (coverpoint == "cp_custom_aqrl"):
       make_custom(test, xlen)
-    elif (coverpoint == "cp_align_byte", "cp_align_word", "cp_align_hword"):
```
jordancarlin, line 1843: Pretty sure you didn't mean to delete this.

bin/testgen.py
```diff
@@ -29,11 +29,18 @@ def insertTemplate(name):
     f.write(f"\n# {name}\n")
     with open(f"{ARCH_VERIF}/templates/testgen/{name}") as h:
         template = h.read()
-    # Replace placeholders with the actual values
+    # Split extension into components based on capital letters
+    ext_parts = re.findall(r'Z[a-z]+|[A-Z]', extension)
+    ext_parts_no_I = [ext for ext in ext_parts if ext != "I"]
+    ISAEXT = f"RV{xlen}I{''.join(ext_parts_no_I)}"
```
jordancarlin, line 35: Doesn't this need to insert underscores between multi-letter extensions? Or is that a known issue you're planning to handle in a later PR?

bin/testgen.py
Marina-Bellido: yeah i did not. idk how that got deleted. let me add it again

bin/testgen.py
Marina-Bellido: After discussing together we decided that we'll need to think more about what extensions should be included in this macro for each extension.

PR #591, opened 2025-05-09
InterruptsS tests at 80% coverage

HamzaYusufJamal:
Had to change a bunch of coverpoints to make them consistent with the idea of "you can't get interrupts in S-mode if mideleg is 0 for that interrupt." Recommended changes to the test plan are in the Interrupts google sheet.

### DIFF
--- a/fcov/priv/InterruptsS_coverage.svh
+++ b/fcov/priv/InterruptsS_coverage.svh
@@ -49,6 +49,9 @@ covergroup InterruptsS_cg with function sample(ins_t ins);
     mstatus_sie_one: coverpoint ins.current.csr[12'h300][1] {
         bins one = {1};
     }
+    mstatus_sie_zero: coverpoint ins.current.csr[12'h300][1] {
+        bins zero = {0};
+    }
     mstatus_tw:  coverpoint ins.current.csr[12'h300][21] {
         // autofill 0/1
     }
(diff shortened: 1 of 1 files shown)


HamzaYusufJamal:
Also found that changes to S-mode always set mstatus.MIE to 0, so this needs to be accounted for in the test plan as well.

davidharrishmc:
I’m not sure what is going on with these changes.  They might be right, but you haven’t explained the reasoning from Friday’s thread.

cp_trigger_m* are removed from InterruptsS S-mode.  You never explained on Friday why they couldn’t be run in that mode.

cp_trigger_sti_* changed from mideleg_ones_zeros to mideleg_ones.  We never got to the root of that discussion.

HamzaYusufJamal:
For the cp_trigger_m*, since mideleg is always 0, sie.M*IE is always read-only zero, so there's no way to raise the enable bit in S-mode. If the enable bit is raised in M-mode, the interrupt just happens before switching modes. The case you had mentioned of "mstatus.MIE = 0, mip.MTIP = 1, mie.MTIE = 0, mideleg.MTI = 0, change to S mode" would run into this same issue, I think. 

For cp_trigger_sti_*, if mideleg is 0, then the same thing happens as with the cp_trigger_m* coverpoints: the only way to raise the interrupt is in M-mode, so we can't do the mode-switch and then raise the interrupt. This is because mideleg being 0 for an interrupt type masks that bit in sie. Therefore, the mideleg=0 case should only be tested in M-mode (which it already is). That amounts to reducing the S-mode case to just mideleg=1s, which is what I've reflected in the covergroup.

davidharrishmc:
For cp_trigger_mti, let’s look at the testplan:
(While in S-mode) With mstatus.MIE = {0/1}, mstatus.SIE = {0/1}, mideleg = {0s/STI+SEI+SSI}, and mie = 1s, use CLINT.MTIMECMP to cause mip.MTIP to rise.
I seem to be missing something. In this situation, the system should take a machine timer interrupt from supervisor mode into machine mode regardless of mstatus.MIE/SIE.  What do you mean by “There is no way to raise the enable bit in S-mode.”  
If mstatus.MIE = 0, mip.MTIP = 1, mie.MTIE = 1 (not zero in Slack), mideleg.MTI = 0, mstatus.MPP = 01, execute MRET, it seems we would go to S-mode, and then there is a pending machine mode interrupt, so the system would trap from S to M mode.  Is something else happening?

PR #595, opened 2025-05-13
Ssstrict trap loop Fix

C0d3ing:
bin/illegalinstrtests.py:

- Excluded ecalls from instructions that generated an ecall. Before this caused a trap loop since an ecall with a random a0 value could cause unwanted behavior (early termination, priv mode changes, or use the return address stored in register ra).

tests/WALLY-init-lib.h:

- Seems like someone accidentally changed this. t0 cannot store the ra value since that would overwrite the instruction size 4/2 (depending on whether the instruction is compressed or not, lines 201 and 204). This is only noticeable for Instruction Access Fault exceptions.


C0d3ing:
tests/WALLY-init-lib.h:

- Added a comment to make it clear that t0 is safe to overwrite for future mantinence

tests/priv/SsstrictInstrS.S:

- Reverted back to fastuncompressed traphandler, forgot to re-add after testing

PR #596, opened 2025-05-14
Commited some coverpoints.

hamza-1821:
cp_cfg_A_off_all
cp_cfg_A_napot_all
cp_cfg_A_napot
cp_cfg_A_na4_all
cp_cfg_A_na4


hamza-1821:
Instead to removing question marks from pmpcfgA_OFF, i have concatenated pmp_hit with it, same for other 3 types.

PR #600, opened 2025-05-17
Made changes to SV48 coverpoints

Zain2050:
Enabled coverage for sv48, which included sv48 in crosses. Most of the changes are just ignoring tera page for sv39.


davidharrishmc:
These look reasonable to me, but I don't have a close handle on the VM coverpoints.  @UmerShahidengr have you reviewed these?

UmerShahidengr:
Yes, these changes are good to go.

PR #607, opened 2025-05-24
Added signature logic for  cp_imm_corners_c.j/c.jal

Marina-Bellido:
CO-AUTHORED Noah and Marina:

Adding signature logic for jumps that are above 2^6 bytes so we don't mess up the alignment. 
Currently work in progress, need to look more at the result signatures


davidharrishmc:
Remove commented out lines that you no longer need
Clean up and remove TODO statements
Replace addi rs1, x0, 1 with li rs1, 1
1473: remove comment about PC that is no longer relevant
There is a fair amount of repetitive code.  Is there a sensible way to share any of it?

PR #609, opened 2025-05-24
odd word alignment for amo.w, atomic endianness

davidharrishmc:


davidharrishmc:
Still needs to be tested.  Need to rebuild tests from riscof so I can run regression-wally --fcov

davidharrishmc:
My environment is too broken to test this, but let's merge it so Marina can test and fix.

templates/testgen/amoadd.w.S
```diff
@@ -0,0 +1,11 @@
+# amoadd.w.S
+# Test odd word-aligned amo operations
+# Put all flavors in here for simplicity
```
jordancarlin, line 3: Comment about all flavors no longer accurate. This is for all of these templates.

PR #611, opened 2025-05-27
Updated makefile to support priv

jacassidy:
Not sure if this is how we want to do this, when trying to simply do a riscv-arch-% was getting errors about the priv extension (ex "EndianM") not being found in rv32/ folder. I have a feeling theres a way to try and catch that so that no error is caused but I'm not sure how to do that


jacassidy:
Just noticed that the instructions say to copy these over manually, I'm concerned having two locations to keep updated with the same code could cause problems especially as more priv vector tests will be written

jordancarlin:
The priv tests will need significant changes to work with riscof. The versions in cvw-arch-verif will not work with riscv-arch-test and the versions in riscv-arch-test will not work with cvw-arch-verif. Since they’re not compatible, there’s no reason to copy them back and forth. I’d recommend new priv tests get written for riscv-arch-test directly and old tests are slowly migrated over as people work on them.

PR #617, opened 2025-05-29
cp_priority and cp_cfg_A_all PMP Coverpoints

sudo-apt-abdullah:
Need a review for these coverpoints and feedback or suggestions for improving their efficiency or coverage quality.


fcov/priv/PMPM_coverage.svh
```diff
@@ -419,7 +419,7 @@ pmpcfgA_OFF: coverpoint {pmpcfg_a,pmp_hit} {
 
 	cp_walk_rs1_pmpaddr: coverpoint ins.current.rs1_val {
 		bins all_zeros = {0};
-		bins all_ones = {-1};
+		bins all_ones = {32'hFFFFFFFF};
```
davidharrishmc, line 422: This works for RV32, but seems it wouldn't work for RV64.  Why not leave -1?

davidharrishmc:
This code is getting complicated and will be challenging for somebody else to understand and maintain.  Please add comments to the subtle parts.
Is there a way to run coverage with this file before you make a PR, so you get the syntax errors out?  It would be worth more internal review first, including a design review with a peer.
This was very time consuming to review. The naming of coverpoint building blocks is not obvious to me and I have to cross-check every one. 
40-43: I don’t understand why there is >> (`G+2).    Please check that your PMP regions are formed correctly, and comment this carefully to explain the logic.  I think it would also help to explain that the standard region is a NAPOT region starting at SAFEREGIONSTART.
46: do you want some `if logic to pick the appropriate standard region based on G?
53: Please expand the input list into many lines and put a comment on each line explaining its meaning.  It looks like you often use lower case to mean the first 16 entries and upper case for the remaining 48, but I can’t tell what pmpaddr vs PMPADDR are doing.    Using capitalization to distinguish variables can be confusing, and is a problem with case-insensitive tools (such as VHDL, if anyone tried dual-language support) – I’d suggest a different convention.
58: I don’t understand how at_region would match an address.  Should it say REGIONSTART rather than STANDARD_REGION?
238: should this be RWXL0001 because it involves 4 bits?
394: This bin named pmpaddr has the same name as a signal.  Confusing.  Also, the meaning of 256 is unclear.
370: Unclear what these dashes are separating, and what the significance of the coverpoints are.  Why are some region 0, some region 1?
385-391: these coverpoints are for xwr, which should be just bits [2:0].  The A bits need to vary for the cp_cfg_L_modify coverpoint.
394: explain why 0x100 is chosen.
397, 401: unclear why this affects pmp_i-1.
405-415: the official names of these registers in the spec don’t have brackets
420-617: why have different walking 1s for pmpaddr vs. pmpcfg?  What if you just had walking_ones, and apply it to all registers you want to test?  It’s harmless writing to the upper bits of pmpaddr and will prove that they don’t do stuff.
623: what is it about this coverpoint that indicates pmpaddr or pmpcfg and justifies write_pmp_csr as the name.  Seems like it is just a csrrw.
774: This coverpoint should check rs1_val, not insn
775: off should be 32 or 64 bits depending on XLEN.
795-828: I don’t understand what the < means here.  If you are trying to make the regions successively larger, it’s completely garbled.
852: how about 0?  More than 32 bits.
861: addr_in_region doesn’t seem to be defined
870: pmp_HIT[k] references 15-63 but the array is declared 47:0 on 855.
879: this is crossed with pmpcfg_xwr, 
882: extra indent.  Comments above appear to apply to this cp even though there is a blank line.
886: should this only apply to RV32?
893-894: why are these bins being ignored?
910-926: should these be inside an ifdef G = 0?

sudo-apt-abdullah:
This code is getting complicated and will be challenging for somebody else to understand and maintain. Please add comments to the subtle parts.
"I've added the comments now."

Is there a way to run coverage with this file before you make a PR, so you get the syntax errors out?
"I always run the coverage with the files before making a PR to confrim there's no syntax errors."

Please check that your PMP regions are formed correctly, and comment this carefully to explain the logic.  I think it would also help to explain that the standard region is a NAPOT region starting at SAFEREGIONSTART.
"I've now defined two macros NON_STANDARD_REGION (NA4 and TOR) and STANDARD_REGION (NAPOT) which will work for all values go G. NON_STANDARD_REGION is simply the PA shifted by 2 while STANDARD_REGION is PA shifted by 2 and has a trailing zero and k = (G-1) trailing ones determinig the size of the region."

Please expand the input list into many lines and put a comment on each line explaining its meaning.
"I've distinguished pmpaddr vs PMPADDR with a different convension also explained in comments."

This bin named pmpaddr has the same name as a signal. Confusing. Also, the meaning of 256 is unclear.
"I've changed the name of coverpoint pmpaddr. 'd256 or 'h100 should be in pmpaddr and we've to attempt to write zero when L = 1. Initially it was 'd7 but you asked to change it to 'h100 so it's not ignored when G is large."

Unclear what these dashes are separating, and what the significance of the coverpoints are. Why are some region 0, some region 1?
"Dashes separates different coverpoints that are simillar like in the case you mentioned, the dashes separates the coverpoints used in cross for cp_cfg_L_modify and cp_cfg_L_modify_TOR. Similarly, the all coverpoints that required in the cross for cp_pmpaddr_walk and cp_pmpcfg_walk are also separated by the dashes. 
Why some region 0 and some region 1 because I need pmpaddr_i and pmpaddr_i-1 for cp_cfg_L_modify_TOR, that's why I used region 1 for cp_cfg_L_modify and region 1 and region 0 for cp_cfg_L_modify_TOR."

These coverpoints are for xwr, which should be just bits [2:0]. The A bits need to vary for the cp_cfg_L_modify coverpoint. 
"Yes, you're right. I've adjusted the field to only rwx." 

The official names of these registers in the spec don’t have brackets
I've added write_to_pmpcfg and write_to_pmpaddre in the cross, so when previous csrrw try to write rs1_val, it will affect pmpaddr_i-1 and pmpcfg_i-1.

"The official names of these registers in the spec don’t have brackets."
Yup, but since I put all register in an array, so pmpaddr[1] implies to pmpaddr1 and pmpcfg[1] to pmpcfg1. But now I've removed the brackets.

Why have different walking 1s for pmpaddr vs. pmpcfg?
"Makes sense, I've removed extra walking ones coverpoint. Now only cp_walk_rs1 will work for both pmpcfg and pmpaddr."

What is it about this coverpoint that indicates pmpaddr or pmpcfg and justifies write_pmp_csr as the name. Seems like it is just a csrrw.
"As a bin, it's just a csrrw, but when it comes in a cross it will specifically try to write pmpaddr or pmpcfg. But now I just changed it to just csrrw."

I don’t understand what the < means here. If you are trying to make the regions successively larger, it’s completely garbled.
"I've tried to build a logic again to make subsequent region twice."

davidharrishmc:
I've reviewed most of this.  Most concerns are resolved but I have two remaining.  Also, line numbering has changed and I can't find/review the original concerns from lines 879 onward.  Can you confirm those last comments were addressed?  Let's get this PR merged soon because it is large and unwieldy!


40: What does 2**(`G-1) do for G = 0? Is special-casing needed? Are these coverpoints tested with both G = 0 and G = 4?
710, 729: these comments seem to assume G = 4.  But the coverpoint should work for any G.
Can you add a comment explaining why val_in_pmpaddr is 256?

hamza-1821:
The ignore bins are for address offsets. The coverpoint addr_offset has 5 bins for different address offsets, that are being used intermittently, in individual coverpoints. For example in cover point cp_cfg_A_napot_all, we need to observe access faults at address, address-4, and just beyond the top of region, but in coverpoint cp_cfg_A_na4, we need to observe access fault at that address, address+4, and just beyong the top of region. Hence ignoring some bins. 
We can get rid of these ignore bins, by declaring seperate coverpoints for each address offset, but that would increase the number of coverpoints. We will have to cross the respective coverpoint with the required address offsets seperately.

sudo-apt-abdullah:
Regarding line 879, you asked about pmpcfg_xwr; I believe it should be included in the cross of cp_cfg_L_modify

cp_cfg_L_modify: cross priv_mode_m, lock_checking, pmp_region, write_pmp_csr, pmpcfg_xwr, val_in_pmpaddr ;
Implies we're in m_mode, region_01 L={1/0}, for three types of region {OFF,TOR,NAPOT}, and we 're trying to write 0 in pmp1cfg and pmpaddr1 while it's initial values are {3'b111} and {32'h100} respectively.

Regarding line 882, I've removed the extra indentation and write the comment to it's related cross which was placed above a blank line.

Line 886 was related to cross I've fixed that as well.

	cp_pmpcfg_walk: cross priv_mode_m, cp_walk_rs1, csrrw, legal_pmpcfg_entries_even ;
	`ifdef XLEN32
		// Will throw illegal instruction when XLEN = 64.
		cp_pmpcfg_zero: cross priv_mode_m, cp_zero_rs1, csrrw, legal_pmpcfg_entries_odd ;
	`endif
	
@hamza-1821 has addressed rest of the concerns related to ignore bins.

Now, I've modified these coverpoints and tested with both G = 0 and G = 4. If G = 0, the region isn't the standard one and PMPADDR has PA shifted by 2. Regarding the coverpoints on line 710 and 729 will work for any value of G.

I've added the comment why 256 is there in the coverpoint.

PR #620, opened 2025-05-30
testgen modification for vector widening instructions

georgiatai:
Modified testgen and a few coverpoint templates to address widening vector instructions. Few notes from the spec regarding the changes:

- When LMUL=2, the vector register group contains vector register v n and vector register v n+1, providing twice the vector length in bits. Instructions specifying an LMUL=2 vector register group with an odd-numbered vector register are reserved.
- A vector register cannot be used to provide source operands with more than one EEW for a single instruction.


georgiatai:
Yes the sigReg part was a mistake. I also made more modifications to testgen so that basic coverpoints for both widen and narrowing instructions work now.

PR #621, opened 2025-05-30
Full coverage for the I extension (fixed I_jal) in rv32 and rv64

Marina-Bellido:
@davidharrishmc 
- In this Pr i properly formatted the changes in pr #607 607.
- Changed auipc for li in signature logic to make it independent from the hardware as proposed in issue 571.
- Full coverage in the I_extension: fixed I_jal and also changed the cp_imm_corners_jal as done in #607 where we allow a range of jumps to allow for signature logic space deviation.


davidharrishmc:
Remove commented out lines that you no longer need
Clean up and remove TODO statements
Replace addi rs1, x0, 1 with li rs1, 1
1473: remove comment about PC that is no longer relevant
There is a fair amount of repetitive code.  Is there a sensible way to share any of it?

Not quite sure how this interacts with your other PR.  If you’re sure it’s ok, I’ll merge.
There is some repetitive code.  Is there any sensible way to share any of it.
1480 says “# temporary fix for bug in compressed branches.”  Can this be resolved?

PR #625, opened 2025-06-02
Fixed vector tests compiling in cvw-arch-verif flow and added Zifence…

davidharrishmc:
…i coverage and Svinval


tests/priv/ZicsrM.S
```diff
@@ -270,6 +270,10 @@ formbits:
     fence.tso
     // Zihintpause
     pause
+    // Svinval
```
jordancarlin, line 273: What about implementations that have Zicsr but not Svinval or Ssstrict? On those implementations this will be unpredictable since they can do anything on an illegal instruction. I think Svinval needs to be its own test file.

testplans/Svinval.csv
```diff
@@ -0,0 +1,4 @@
+Instruction,Type,RV32,RV64,cp_asm_count,cp_custom
```
jordancarlin, line 1: Won’t this generate the Svinval coverpoints in the unpriv directory? Svinval is a privileged extension, so that is kind of misleading. Especially since it’s just using a custom template, we might be better off just writing the covergroup in the priv directory and skipping the generation for it.

templates/testgen/fence.i.S
```diff
@@ -0,0 +1,20 @@
...
+# custom tests for fence.i
+
+# cp_custom_fencei
+
+    li t2, 0        # running total
+    jal selfmodify  # should increment t2 to 1
+    la t0, selfmodify
+    lw t1, 8(t0)    # load the addi t2, t2, 5 statement
+    sw t1, 0(t0)    # store it at selfmodify
+    fence.i         # synchronize self-modified code so j done is replaced with ret
```
jordancarlin, line 11: Comment incorrect. This changes the addi. It doesn’t replace j done with ret.

bin/testgen.py
```diff
@@ -2400,9 +2400,9 @@ def write_tests(coverpoints, test, xlen=None, vlen=None, sew=None, vlmax=None, v
       make_sbox(test, xlen)
     elif (coverpoint in ["cp_sc", "cp_prev_lr", "cp_prev_sc", "cp_custom_sc_after_sc", "cp_custom_sc_after_store",
                          "cp_custom_sc_after_load", "cp_sc_fail", "cp_address_difference", "cp_custom_sc_lrsc",
-                         "cp_custom_sc_addresses", "cp_custom_rd_corners"]):
+                         "cp_custom_sc_addresses", "cp_custom_rd_corners", "cp_custom_svinval"]):
```
jordancarlin, line 2403: `cp_custom_svinval` no longer needed now that it's a priv test.

tests/priv/ZicsrM.S
```diff
@@ -270,6 +270,10 @@ formbits:
...
+    sinval.vma x0, x0
+    sfence.w.inval
+    sfence.inval.ir
```
jordancarlin, line 276: These tests are now duplicated in both `Svinval.S` and here in `ZicsrM.S`. They should be removed from this file and just be in `Svinval.S`.

fcov/priv/Svinval_coverage.svh
```diff
@@ -0,0 +1,59 @@
...
+    option.per_instance = 0;
+    cp_asm_count : coverpoint ins.ins_str == "sinval.vma"  iff (ins.trap == 0 )  {
+        //Number of times instruction is executed
+        bins count[]  = {1};
+    }
+endgroup
+// ---------------------
+function void svinval_sample(int hart, int issue, ins_t ins);
+
+    case (traceDataQ[hart][issue][0].inst_name)
```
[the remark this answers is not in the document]
davidharrishmc: I don't think it's harmful and it night be slightly more efficient?

jordancarlin:
Should we go one step further and do a full cross of priv mode, TVM, and asm_count to make sure TVM doesn't impact the instruction in M mode? It's a minimal number of additional instructions and seems like it would provide better coverage. We also need a spreadsheet testplan for the Svinval extension.

davidharrishmc:
Good suggestions.  Testplan generated and full cross running with full coverage.

PR #627, opened 2025-06-02
Floating point signature but reduced coverage 

Noah-Fotenos:
Added signatures to ZfaD,  changed ISA ordering in testgen for spike compatibility, reordered coverpoint for test gen register instantiation bug, and fixed bug in offset macros to align offset with diff regwidths.


bin/testgen.py
```diff
@@ -138,18 +148,23 @@ def writeSIGUPD(rd):
     return l
 
 def writeSIGUPD_F(rd):
-    # *** write this
-    return ""
+    global sigupd_count  # Allow modification of global variable
+    sigupd_count += 2  # Increment counter by 2 on each call since SIGUPD_F macro stores two registers to memory
```
jordancarlin, line 152: You might need a separate counter for floating point signature entries vs integer entries. The signature region is currently xlen*sigupd_count, but floating point entries in the signature will take up flen bits, not xlen bits. This would be a problem on rv32gc when flen is longer than xlen.

bin/testgen.py
```diff
@@ -21,19 +21,29 @@
...
 def insertTemplate(name):
+    global sigupd_count
     f.write(f"\n# {name}\n")
     with open(f"{ARCH_VERIF}/templates/testgen/{name}") as h:
         template = h.read()
     # Split extension into components based on capital letters
     ext_parts = re.findall(r'Z[a-z]+|[A-Z]', extension)
     ext_parts_no_I = [ext for ext in ext_parts if ext != "I"]
+    if 'D' in ext_parts_no_I:
+      ext_parts_no_I = ['D']+ext_parts_no_I
```
jordancarlin, line 32: Won’t this end up with D in the list twice?

bin/testgen.py
```diff
@@ -21,19 +21,29 @@
...
+      sigupd_count = sigupd_count * 2 # double the space needed for double precision
```
jordancarlin, line 33: For rv64 this will be double the necessary signature size. See my other comment for a suggestion to avoid this using 2 counters.

PR #629, opened 2025-06-03
Modified testgen for vsext/vzext test generating and fixed covergroupgen

georgiatai:
- Fixed covergroupgen such that if an instruction is not available for certain SEWs, the correct covergroups are generated.
- Modified testgen for vsext/vzext instructions to run on easy coverpoints.

Note: this PR may have merging issues as I created the branch from the previous PR, but nothing from the previous PR should be modified


georgiatai:
Added modifications for vmacc instructions too. They were mistakenly placed into masked instruction types when they should have been normal integer instructions.

georgiatai:
Opened a new PR (#631) that includes this PR and more vector instructions

PR #638, opened 2025-06-04
Separate vector and non-vector testgen

georgiatai:
Separated vector and non-vector testgen into two files. Changed Makefile for this change too.


bin/vector-testgen.py
```diff
@@ -0,0 +1,1212 @@
...
+                0xC93A]  # -10.4531 random negative
+
+  # fcornersQ = [] # TODO: Fill out quad precision F corners
+
+# generate files for each test
+  for xlen in xlens:
+    # for E_ext in [False, True]:
+    for E_ext in [False]: # for testing only ***
+      if (E_ext):
+        extensions = ["E", "M", "Zca", "Zcb", "Zba", "Zbb", "Zbs"]
```
davidharrishmc, line 1081: Can you trim more stuff that isn't needed for vector, such as these non-vector extensions and the E extension?

PR #643, opened 2025-06-04
updating c.sw to use the same signature logic a I_sw

Marina-Bellido:
This function also updated Zcd_c.sfd,    Zca_c.sd.S,       Zcf_c.fsw. I'll be checking them all tomorrow. So there is no coverage issue but they have no errors. Those instructions probably need other changes too, so it is fine to merge this pr

### DIFF
--- a/bin/testgen.py
+++ b/bin/testgen.py
@@ -891,25 +891,35 @@ def writeHazardVector(desc, rs1a, rs2a, rda, rs1b, rs2b, rdb, testb, immvala, im
     hazardLabel += 1
 
   elif insMap[instype].get('loadstore', 0) == 'store':
+
     lines = lines + "mv x" + str(rs1b) + ", x" + str(sigReg) + " # move sigreg value into rs1\n"
     sigReg = rs1b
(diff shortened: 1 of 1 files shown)


Marina-Bellido:
**Rethinking this to make it more intuitive for the user:
Example of new setup for the custom test**

# sw.S
# Test word-aligned sw operations
    mv t0, SIGPOINTER #move signature pointer to t0.
    li t2, 0x12345678
    li t3, 0x9ABCDEF0

    # cp_align_word: imm[2]=0 (aligned store)
    sw t2, 0(t0)
    sw t3, 4(t0)
    #Increment after since we are storing two words using offset 0 and 4
    SIG_POINTER_INCREMENT(2)
    
    addi t0, t0, REGINC   -------> testgen replace by: (REGWITH *2)


**SIG_POINTER_INCREMENT(X) -> The number inside will determine by how much we multiply REGWIDTH and also increment sigupd_count. This way the user does not have to worry about making the signature region big enough, just incrementing the offset every time she creates an instruction.**

PR #645, opened 2025-06-05
First draft for vector corner value coverpoints/tests and updating VLEN

georgiatai:
- updating VLEN to 512 instead of 256 since sail is using 512
- first draft of vector corner value coverpoints and tests
-- collecting coverage, not 100% because rvviTrace is having issues updating the vector values and thus the coverpoints are not recognizing their values (confirmed in sail trace file that the vectors are getting correct values)

### DIFF
--- a/bin/covergroupgen.py
+++ b/bin/covergroupgen.py
@@ -12,6 +12,7 @@ import os
 import csv
 import re
 import sys
+import math
 
 ##################################
 # Functions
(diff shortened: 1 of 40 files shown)


georgiatai:
PR #648 also includes fixing errors on sail parsing, which I'm not sure would cause a merging issue or not later. But #648 is more urgent so please review that first. Thanks!

PR #647, opened 2025-06-06
adding signatures to more ftype instructions and bug fix 

Noah-Fotenos:
Some clean up
Changes to ensure no conflicts with stack pointer when doing stores,
Adding FCSR to signatures for other Ftype instructions that don't use SIGUPD_F,


davidharrishmc:
Should this file also have the testplan per the commit summary?

PR #657, opened 2025-06-10
fixed broken coverpoints

jacassidy:


templates/coverage/priv/cp_exceptionsv_mstatus_vs_off.txt
```diff
@@ -3,7 +3,7 @@
     }
 
     vsstatus_vs_active : coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vsstatus", "vs") {
-        bins active = {1};
+        ignore_bins active = {0};
     }
```
davidharrishmc, line 7: Should 0 be OFF?

jacassidy: yes, we want to test every case except off: note "ignore_bins"

jacassidy: should I make the other bins more verbose?

jacassidy:
Ended up doing the bins more verbose. I did not know that was how the ignore_bins naming convention worked!

PR #673, opened 2025-06-14
Separate Coverage files for PMPF, PMPZaamo and PMPZalrsc

sudo-apt-abdullah:
Most of the stuff is same, but we're asked to keep separate files for all different modes and extensions mentioned in testplan. If anything can be improved, please let me know.


fcov/priv/PMPF_coverage.svh
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	standard_region: coverpoint ins.current.csr[12'h3B0] {
+		bins standard_region = {`STANDARD_REGION};
+	}
+
+	read_fp_instr: coverpoint ins.current.insn {
+		wildcard bins flh = {32'b????????????_?????_001_?????_0000111};
+		wildcard bins flw = {32'b????????????_?????_010_?????_0000111};
+		`ifdef XLEN64
```
davidharrishmc, line 59: double precision floating-point doesn't depend on XLEN, so this can be dropped.  We could separate out flh and fld into PMPZfh and PMPD files, but I'm ok testing all three of them here if Umer concurs.

PR #675, opened 2025-06-16
Vector Load store coverpoints

jacassidy:


jacassidy:
I would write a comment, but Im not sure anything would be adequate

jacassidy:
Noone has had time to review yet in person, we can do one tomorrow

jacassidy:
@rosethompson that would be great! Is there a time today or tomorrow that works best for you PST?

templates/coverage/cp_vs3_lte24.txt
```diff
@@ -0,0 +1,10 @@
+    cp_vs3 : coverpoint ins.get_vr_reg(ins.current.vs3)  iff (ins.trap == 0 )  {
+        //VD register assignment excliding bins
```
davidharrishmc, line 2: excluding - maybe there is a way to spell check?

jacassidy: I have installed spellcheck and fixed this instance

PR #676, opened 2025-06-17
Refining vector length suite

georgiatai:
Last week, we changed the definition of some coverpoints in length suite:
- cr_vl_lmul: Cross all legal LMUL = 1/2/4/8/{f2/f4/f8} and vl = 1, 2<=random<VLMAX, VLMAX
- cr_vtype_agnostic: Cross vta/vma
- cp_masking_corners: deleted checkerboard
P.S. variables not specified should be random, including mask/no mask and what the mask value would be

Some issues fixed:
- what LMULs are legal vary across SEWs
  - defined macros to see which LMULs are legal architecturally
  - cr_vl_lmul needs different coverpoint templates for different SEWs, modified covergroupgen for this
  - added ifdefs around fractional LMUL tests
    - SEWMIN and ELEN are included in the header of the test currently as I have not figured out how to pass them in with -D yet, but they should be changed later
- VLEN cannot be passed into testgen since the tested should be pre-generated for users
  - cases where VLMAX are calculated are redesigned so that VLMAX can be obtained in assembly, e.g. mask corner values
  - when vl is not a fixed number (vl = VLMAX/random), then set signature offset avl to 2048 (max VLEN/VLMAX)
- widening instructions do not exist on ELEN suites
  - added ifdefs to include them, both in covergroup and tests
- redesigned corner values for mask logical instructions (eew = 1) since it doesn't make sense to only test 1 element in these cases. vl is set to 8 as it is the minimum VLMAX in any cases

### DIFF
--- a/templates/coverage/cr_vl_lmul_sew8.txt
+++ b/templates/coverage/cr_vl_lmul_sew8.txt
@@ -1,13 +1,13 @@
     cp_csr_vtype_lmul_all_sew8 : coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vlmul")  iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vsew") == 0) {
         // Value of VTYPE.vlmul (vector register grouping), SEW = 8
         `ifdef LMULf8_SUPPORTED
-        bins eigth  = {5};
+            bins eigth  = {5};
         `endif
         `ifdef LMULf4_SUPPORTED
-        bins fourth = {6};
+            bins fourth = {6};
         `endif
         `ifdef LMULf2_SUPPORTED
-        bins half   = {7};
+            bins half   = {7};
         `endif
         bins one    = {0};
         bins two    = {1};
--- a/templates/coverage/cr_vl_lmul_lmul4max_sew16.txt
+++ b/templates/coverage/cr_vl_lmul_lmul4max_sew16.txt
@@ -1,10 +1,10 @@
     cp_csr_vtype_lmul_all_lmul4max_sew16 : coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vlmul")  iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vsew") == 1) {
         // Value of VTYPE.vlmul (vector register grouping), SEW = 16, excluding LMUL = 8
         `ifdef LMULf4_SUPPORTED
-        bins fourth = {6};
+            bins fourth = {6};
         `endif
         `ifdef LMULf2_SUPPORTED
-        bins half   = {7};
+            bins half   = {7};
         `endif
         bins one    = {0};
         bins two    = {1};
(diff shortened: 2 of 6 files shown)


templates/testgen/testgen_header_vector.S
```diff
@@ -27,3 +27,30 @@ RVTEST_CASE(0,"TestCase",Instruction)
 RVTEST_V_ENABLE()
 RVTEST_VALBASEUPD(DATA_BASE, vector_data)
 RVTEST_SIGBASE(SIG_BASE, signature_base)
+
+#define SEWMIN 8
+#define ELEN 64
```
jordancarlin, line 32: Is the idea for these to be passed in by RISCOF when compiling?

georgiatai: Yes, I’m still looking for changes that have to be made to pass it in -D. They will be removed later

templates/coverage/cr_vl_lmul_sew8.txt
```diff
@@ -0,0 +1,32 @@
+    //////////////////////////////////////////////////////////////////////////////////
+    // cr_vl_lmul_sew8
+    //////////////////////////////////////////////////////////////////////////////////
+
+    cp_csr_vtype_lmul_all_sew8_lmul_le_8 : coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vlmul")  iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vsew") == 0) {
+        // Value of VTYPE.vlmul (vector register grouping), SEW = 8
+        `ifdef LMULf8_SUPPORTED
+        bins eigth  = {5};
```
jordancarlin, line 8: Indent inside the `ifdef. Global comment that applies to lots of these files.

jordancarlin (commented):
`vector-testgen` is getting hard to follow, but I think we should keep it as is for now and then do a refactor once there is a more complete picture of everything it needs to do. Will definitely need to be reworked before upstreaming to riscv-arch-test though because I doubt many people will be able to understand it in its current state.

templates/coverage/covergroup_sample_vector_widen.txt
```diff
@@ -0,0 +1,5 @@
+        `ifndef ELENEFFEW
```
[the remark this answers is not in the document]
georgiatai: It's supposed to be ifndef because widening instructions don't exist when SEW=ELEN, and ELEN is defined by the largest Vx{SEW}_COVERAGE in config. Anything that is above ELEN should be neglected by this coverage macro, and thus the ifndef choice

templates/coverage/cr_vl_lmul_sew8.txt
```diff
@@ -0,0 +1,32 @@
...
+        bins eight  = {3};
+    }
+
+    cp_csr_vl_corners : coverpoint vl_check(ins.hart, ins.issue)  iff (ins.trap == 0 )  {
+        // Corners values of VL (vector length)
+        ignore_bins vl_zero   = {vl_zero};
+        ignore_bins vl_illegal = {vl_illegal};
+    }
+
+    cr_vl_lmul_sew8 : cross cp_csr_vtype_lmul_all_sew8_lmul_le_8, cp_csr_vl_corners  iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vsew") == 0)  {
```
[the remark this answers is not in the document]
jordancarlin: @jacassidy this should be addressed in your PR

PR #690, opened 2025-06-23
Zcf extension fixes + cleanup

Noah-Fotenos:
ensure no sp load conflicts with sp instructions

c.flwsp needs to load to rs2 not rd.

comment cleanup

c.flw is a float instruction so writeTest puts the right SIGUPD_F

fix incrementing to byte size


davidharrishmc:
merge conflict

PR #714, opened 2025-07-03
fixed bug in the testbench.sv

Marina-Bellido:
The ExceptionsM.trace prints the mode in decimal "MODE 3", but the testbench was interpreting that in binary so I changed it to %d. This fixed the bug and now priv_mode_m outputs 11 (before it was 00), and also went from 57% to 100% coverage in ExceptionsM.


davidharrishmc:
Whoo!!!!

PR #716, opened 2025-07-05
Fixed cp_grain coverpoints

hamza-1821:
The testplan for cp_grain_preserved is pretty much same as cp_grain in case of pmpcfg.A=NAPOT. since changing pmpcfg.A and reading back pmpaddr0 will be a part of the test, is there a need to create a cross for cp_grain_preserved?


davidharrishmc:
Deleted pmp_grain_preserved from the testplan.

hamza-1821:
The recent commit implements the writing and reading in a different mode. CSRR is current and CSRRW is prev instruction.

hamza-1821:
After the recent changes, and addressing your feedback, it boils down to only checking ins.prev.csr[12'h3A0] and ins.current.csr[12'h3A0] in the coverage report.

PR #723, opened 2025-07-07
Adding instruction cillegal to disassembly logic

Marina-Bellido:
My transcript was not disassemblying the cillegal instructions and just passing the previous instruction. 
So added code in the disassembler to be able to decode c_illegal.


jordancarlin:
I have concerns with this since it modifies an autogenerated file. @Marina-Bellido Did this matter for coverage or was it just a debugging/display aid? We should definitely improve it to give proper disassembly either way, but it impacts what the best solution is.

PR #724, opened 2025-07-07
vector sigupd macro update in cvw-arch-verif

georgiatai:
fixes errors for make --jobs


jordancarlin:
If these files always need to be kept in sync, would we be better off just making it a symlink?

georgiatai:
This may be a good idea, but I'm not 100% sure the reason it was kept separate originally, and would like to get some input from @Marina-Bellido @Noah-Fotenos.

PR #732, opened 2025-07-14
Move integer corner values to central file

jordancarlin:
The same corner values were maintained in coverage templates, testgen, and now the CTP. This is an attempt to centralize the corner values in one location and populate coverpoints, tests, and documentation from the same source.

I started with just the rs1/rs2 integer corner values as a proof of concept, but this can be extended to all of the corner values (floating point, immediates, etc.) so they aren't repeated in multiple places.

This version moves the corner values into a CSV file. This CSV is directly included in the CTP to populate the relevant table. A new python script reads this CSV and generates the values for tests and for coverpoint bins. The other approach we could take is to leave the coverpoint templates unchanged and have testgen directly read the bins from the coverpoints, but that still requires the data to be replicated in the CTP and the parsing logic would be more complicated, so I went with this approach as a first attempt.

Also remove unused rd_corners from testgen.


jordancarlin:
@davidharrishmc I opened this as a draft to get initial thoughts. If you like the direction then I will get it cleaned up and ready for merging.

davidharrishmc:
I'm ok with this if you feel it is important.  The corners haven't really changed since they were written, and updating them isn't too hard, so I don't know that automating it is a priority.  On the other hand, the work is already done.

davidharrishmc:
ok to close?

PR #738, opened 2025-07-16
Better Python setup

Timmmm:
As discussed with @jordancarlin, this improves the Python setup in various ways:

1. Use `Pyproject.toml` and put the code in an actual package. This has two big benefits: a) imports work sanely, b) you can use third party dependencies sanely. Note the "sanely" part relies on using `uv`.
2. Enable type checking with Pyright in CI. I fixed the types in `covergroupgen.py` to demonstrate this (see the commit for more details).
3. Use Typer for `vector_coverage.py`. This is just a much nicer (and better typed) way of doing CLI interfaces. It adds fancy backtraces too. I did have to simplify the `-j` interface a bit (`-j 4` instead of `-j 4` and `-j 0` instead of `-j`).

There are a couple of other things I would do to improve things further:

1. Enable `ruff` linting and formatting in pre-commit (@jordancarlin is on this already).
2. When the Python code loads data files it should use `pkgutils` to do so, so that you can distribute the whole thing as a Python package. I'll have to remember how to actually do this...

I really really really really really strongly recommend you adopt all of these best practices before it is too late. There are already several thousand lines of Python and it's *MUCH* harder to fix all this in hindsight than to do it right from the start.


Timmmm:
I've signed the CLA. What do you think of this PR? Shall I split it out into separate ones so you can merge them?

jordancarlin:
> I've signed the CLA. What do you think of this PR? Shall I split it out into separate ones so you can merge them?

Great. Thanks @Timmmm. We're still sorting out how we want to proceed with the Python package, but I definitely want to enable type checking sooner rather than later, so it would be great if you could at least split the `covergroupgen.py` updates out into a separate PR to start with. I enabled the Ruff linter recently, so there might be significant merge conflicts.

PR #756, opened 2025-07-21
add rv64 to combine .S with comments and clean up in makegen

Noah-Fotenos:


jordancarlin (changes requested):
Good to remove the duplicate files, but right now this breaks the existing Makefile for compiling and running the tests without RISCOF on Wally. While the eventual goal is to move to this new approach, we can't break the existing approach until we are ready to fully switch over. Please figure out a way to make both approaches work with a single set of header files.

The two versions of the header files were also different. Please confirm with the rest of the team that the version you are keeping here is the correct version. I don't want work that has been done by others getting deleted by mistake.

Noah-Fotenos:
Reverted the file changes. I think we'll have to keep copies until self-check is done.

PR #764, opened 2025-07-30
Modified VM_Exceptions covergroups

Zain2050:
**Changes in ExceptionsVM Zaamo & Zalrsc:**
I have removed `d_page_table_entry_bad` & `d_phys_address` from VMZaamo & VMZalrsc covergroups as they are not involved in any cross and are being covered in the main ExceptionVM covergroup.

**Changes in ExceptionsVM:**

`misaligned_pagespanning_address` is not required. It is rs1+imm which is actually our virtrual address. The last 12 bits of VA and PA are same. However, the cross is trying to check (rs1+imm = 12'b111111111110) with (phys_addr = 12'b111111111110 & 12'b0), which is not possible. I can add an ignore, but I think we don't need to check rs1+imm.

`cp_instr_page_fault_m, cp_misaligned_priority_fetch_m & cp_medeleg_fetch_m` are not possible as SUM does not affect instruction fetches. Instruction fetches always use physical address in M mode, and therefore we can't have PTEs and instr page faults while using physical address.

I have changed `jalr: coverpoint ins.current.insn` to `jalr: coverpoint ins.prev.insn`. Here is the log from a test I wrote for it.

Here, jalr is executed but pte, va & pa values are not yet right (see line beginning with -->)
```
# --> disass = 000780e7 jalr x1, 0(x15), inst = 000780e7, va = 00000300800015e0, phy_i = 000000800015e0, pte_i = 00000000000000cb
# Info 3816: 'refRoot/cpu', 0x00000300800015e0(main+de0): Supervisor 000780e7 jalr    x1,0(x15)

# Info   MEMX 0x300800015e0 0x800015e0 2 80e7
# Info   MEMX 0x300800015e2 0x800015e2 2 0007
# Info   x1 0000030080001324 -> 00000300800015e4
```
We perform page table walk, and get the correct values, but inst changes to 0000.
```
# --> disass = 0000 c.illegal, inst = 00000000, va = 0000040230044ffe, phy_i = 00000080005ffe, pte_i = 0000000020001400
# Info 3817: 'refRoot/cpu', 0x0000040230044ffe: Supervisor *** FETCH EXCEPTION ***
```
Our desired values appear when it tries to jump to the calculated PA, but inst changes to 0000. Therefore, we should use prev instead of current.

### DIFF
--- a/fcov/priv/ExceptionsVMZaamo_coverage.svh
+++ b/fcov/priv/ExceptionsVMZaamo_coverage.svh
@@ -21,7 +21,7 @@
 // and limitations under the License.
 ////////////////////////////////////////////////////////////////////////////////////////////////
 
-`define COVER_EXCEPTIONSVM
+`define COVER_EXCEPTIONSVMZAAMO
 covergroup ExceptionsVMZaamo_cg with function sample(ins_t ins);
     option.per_instance = 0;
     `include "coverage/RISCV_coverage_standard_coverpoints.svh"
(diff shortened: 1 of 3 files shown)


davidharrishmc:
Zain will check whether jalr should fault at jalr or at the destination.

davidharrishmc:
Waiting on Zain to modify fetch tests.  Then ready to merge.

Zain2050:
I have modified the mprv fetch coverpoints. I have also made a minor change of getting **prev** value of **MPP** instead of the current one. In case of a trap the value of MPP change to 3. It has the same reason as using prev for priv_mode. The other VM covergroups written by Huda are also using mpp.prev

PR #765, opened 2025-07-31
CTP updates

davidharrishmc:
Flush out extensions, hypervisor normative statements, edge cases


docs/ctp/ctp.adoc
```diff
@@ -385,14 +473,14 @@ and the architectural state of the hart.
...
+Every bin in every coverpoint is associated with a specific test footnote:[cp_asm_test is an exception, just counting the number of times the instruction was executed]. Unprivileged tests sweep some feature under test (such as the destination register rd) while randomizing all applicable register IDs and source values (including values to be loaded from memory).  For example, the unprivileged test for the cp_rd coverpoint of the add instruction contains 32 add instructions, using the 32 different choices of rd.  The rs1 and rs2 register numbers and values are selected randomly. <<Appendix A: Examples>> gives examples of unprivileged coverpoints and tests meeting these requirements and the coverpoints of <<t-I-coverpoints>> for the `add` instruction.  Each test is self-checking, checking the destination register value against one provided by a reference model.  Certain tests also test other state such as stores or floating-point flags, as described in those sections.
```
jordancarlin, line 483:
```suggestion
Every bin in every coverpoint is associated with a specific test footnote:[cp_asm_count is an exception, just counting the number of times the instruction was executed]. Unprivileged tests sweep some feature under test (such as the destination register rd) while randomizing all applicable register IDs and source values (including values to be loaded from memory).  For example, the unprivileged test for the cp_rd coverpoint of the add instruction contains 32 add instructions, using the 32 different choices of rd.  The rs1 and rs2 register numbers and values are selected randomly. <<Appendix A: Examples>> gives examples of unprivileged coverpoints and tests meeting these requirements and the coverpoints of <<t-I-coverpoints>> for the `add` instruction.  Each test is self-checking, checking the destination register value against one provided by a reference model.  Certain tests also test other state such as stores or floating-point flags, as described in those sections.
```

docs/ctp/ctp.adoc
```diff
@@ -1860,9 +1948,33 @@ environment.|
...
 
-*** includes mret/sret
+Zicsr privileged tests access to CSRs that are mandatory in each privilege mode.
+It also tests mandatory privileged instructions: `ecall`, `ebreak`, `mret`, and `sret`.
+The CSR read/write/set/clear instructions themselves are tested in the
+unprivileged Zicsr tests of <<Zicsr Control/Status Register Instructions>>.
+
+`wfi` is tested in <<Interrupts>> and `sfence.vma` is tested in <<Sv* Virtual Memory>>.
+
+Zicsr does not test reserved or custom instructors or CSRs because their behavior is Unspecified.
```
jordancarlin, line 1975:
```suggestion
Zicsr does not test reserved or custom instructions or CSRs because their behavior is Unspecified.
```

PR #770, opened 2025-08-07
Started trickbox

davidharrishmc:
Started Trick Box section of CTP


docs/ctp/src/v.adoc
```diff
@@ -295,3 +295,15 @@ There are several Zvk* vector cryptography extensions.  Zvks ShangMi extensions
...
+==== ZfaZvfh Half-Precision Load Immediate
+
+When Zfa and Zvfh are both implemented, RISC-V provides a fli.h instruction (<<t-ZfaZvfh-coverpoints>>) even if full Zfh is not supported.
+
+[[t-ZfaZvfh-coverpoints]]
+.ZfaZvfh Instruction Coverpoints
+[options=header]
+[%AUTOWIDTH]
+,===
+include::../../testplans/ZfaZvfh.csv[]
```
jordancarlin, line 308:
```suggestion
include::{testplansdir}/ZfaZvfh.csv[]
```

docs/ctp/trickbox.adoc
jordancarlin, line 1: Needs to move to src directory.

docs/ctp/a.adoc
jordancarlin, line 1: Needs to move to src directory and update test plan paths to use `{testplansdir}`.

docs/ctp/src/exceptions.adoc
```diff
@@ -1,28 +1,173 @@
...
 ==== ExceptionsM
 
+These tests are run on any configuration with machine mode that supports exceptions.
+
+[[t-ExceptionsM-coverpoints]]
+.ExceptionsM Coverpoints
+[options=header]
+[%AUTOWIDTH]
+,===
+include::{testplansdir}/ExceptionsM.csv[]
```
jordancarlin, line 15: Might be worth commenting out all of these nonexistent includes until the CSV exists so we don’t get a bunch of errors when building.

docs/ctp/src/aia.adoc
```diff
@@ -0,0 +1,24 @@
+
+=== AIA Advanced Interrupt Architecture
+
+This section describes AIA extension test plans.
```
jordancarlin, line 4: This probably needs a non-isa test plan in addition to the 2 isa extensions.

docs/ctp/src/e.adoc
jordancarlin, line 1: Are you envisioning separate testplan CSV files for the E variants of extensions? If so, what do you think will be different? It seems like they will have the same coverpoints and the same instructions.

davidharrishmc: The tests are definitely different.  I thought it would be clearer to have testplans aligned to the tests even though they are identical to I except what get_gpr_reg returns.

Any better ideas?

docs/ctp/src/aia.adoc
davidharrishmc: I think you are right.  Added a first cut.  Likely more as we understand it better.

docs/ctp/src/privmisc23.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+[[t-Sscofpmf-coverpoints]]
+.Sscofpmf Coverpoints
+[options=header]
+[%AUTOWIDTH]
+,===
+//include::{testplansdir}/Sscofpmf.csv[]
+,===
+
+==== Smstateen Machine State Enable
```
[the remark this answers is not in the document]
davidharrishmc: ok

PR #771, opened 2025-08-07
adjusted coverpoints for tor non overlap in PMPM

hamza-1821:
For coverpoint cp_cfg_A_tor_nonoverlap we have 3 test cases each time with a different pmpaddr0 value. Since the region is locked, so I have created 3 seperate test files to implement 3 test cases. The coverpoint definition is updated for each of those test cases named as 1, 2 and 3.

### DIFF
--- a/fcov/priv/PMPM_coverage.svh
+++ b/fcov/priv/PMPM_coverage.svh
@@ -421,14 +421,20 @@ covergroup PMPM_cg with function sample(
 		bins pmpaddr1 = {`REGIONSTART+`g};
 	}
 
-	pmp_addr_for_tor_nonoverlap: coverpoint {pmpaddr[1], pmpaddr[0]} { // pmpaddr0 >= pmpaddr1.
-  		bins range1 = {`NON_STANDARD_REGION, `NON_STANDARD_REGION};
-		bins range2 = {`NON_STANDARD_REGION, `NON_STANDARD_REGION+`g};
-		bins range3 = {`NON_STANDARD_REGION, {$bits(pmpaddr[0]){1'b1}}}; //pmpaddr0 = all 1s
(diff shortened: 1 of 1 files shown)


fcov/priv/PMPM_coverage.svh
```diff
@@ -421,14 +421,20 @@ covergroup PMPM_cg with function sample(
 		bins pmpaddr1 = {`REGIONSTART+`g};
 	}
 
-	pmp_addr_for_tor_nonoverlap: coverpoint {pmpaddr[1], pmpaddr[0]} { // pmpaddr0 >= pmpaddr1.
-  		bins range1 = {`NON_STANDARD_REGION, `NON_STANDARD_REGION};
-		bins range2 = {`NON_STANDARD_REGION, `NON_STANDARD_REGION+`g};
-		bins range3 = {`NON_STANDARD_REGION, {$bits(pmpaddr[0]){1'b1}}}; //pmpaddr0 = all 1s
+	pmp_addr_for_tor_nonoverlap1: coverpoint ({pmpaddr[1], pmpaddr[0]}=={`NON_STANDARD_REGION, `REGIONSTART>>2}) { // pmpaddr0 >= pmpaddr1.
```
davidharrishmc, line 424: range1 had been `NON_STANDARD_REGION, `NON_STANDARD_REGION and now non overlap1 is `NON_STANDARD_REGION, `REGIONSTART>>2.  Is this intentional to change?

hamza-1821:
> I understand using three tests. Why is it necessary to have three different coverpoints, rather than one coverpoint with three bins? Can't each bin be hit by a different test?

In a single coverpoint for pmpaddr, the individual bins were being hit, but the cross coverage was 0. I tried adding assertions in the bin definition to see if the bins hit at the same point, and they do. So I broke the coverpoint into 3 seperate ones, each having their own assertion, this made it easy to keep track of the signal, and also the coverage of the cross was 100%.

PR #780, opened 2025-08-15
Modified existing CBO_VM coverpoints & added new ones as well

Zain2050:
I have added some new coverpoints which check for store access faults. Therefore, I have removed Mcause coverpoint and made 2 separate coverpoints for store page fault and store access fault.
I have also added coverpoints to check for DAU faults for non leaf PTEs when zicbom instructions are executed. 

**Modifications to existing coverpoints:**
I have removed **write_acc** because it does not tell us that a cbo instruction is being executed. Even in RTL, CMOpM is used instead of WriteAccess. 
Coverpoints for misaligned_PPN were incorrect. I have made the same change that I made for RV64VM misaligned coverpoint. 

Coverpoints **spage_nowrite_s_d** and **upage_umode_nowrite_u** were wrong. It was checking for store page faults when W=0 (same is written in the test plan). But Zicbom instructions do not depend upon W permission. As mentioned in section 19.4.2.2 of Unpriv Spec
```
A cache-block management instruction is permitted to access the specified cache block whenever a
load instruction or store instruction is permitted to access the corresponding physical addresses.
```
Therefore, as we have R=1, it won't fault. Either we can use this test case to confirm that cbo doesn't fault when W=0 and R=1. Or we could update the coverpoint to check for fault when R=0 and W=0.

Similarly, it won't fault when D=0. In the same section of spec we have,

```
During address translation, the instruction also checks the accessed bit and may either
raise an exception or set the bit as required.
```
```
As implied by omission, a cache-block management instruction does not check the
dirty bit and neither raises an exception nor sets the bit.
```

The **testplan** also needs to be updated accordingly.

### DIFF
--- a/fcov/rv32_priv/RV32CBO_VM_coverage.svh
+++ b/fcov/rv32_priv/RV32CBO_VM_coverage.svh
@@ -70,6 +70,16 @@ covergroup RV32CBO_VM_exceptions_cg with function sample(ins_t ins);
         wildcard bins leaflvl_s = {8'b???01111};
     }
 
+    pointer_PTE_d: coverpoint ins.current.pte_d[7:0] {
+        wildcard bins pointer = {8'b00?00001};
+    }
+
+    PTE_DAU_d: coverpoint ins.current.pte_d[7:0] {
+        wildcard bins nonleaf_D_bit = {8'b1?0?0001};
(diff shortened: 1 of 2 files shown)


Zain2050:
These coverpoints were linked to the testplan with //exp comments. I have removed these comments from the coverage files and replaced exp with coverpoint names in the testplan for easier review.

PR #782, opened 2025-08-19
CTP improvements

davidharrishmc:
Removed RV32/RV64 where possible from coverage files.
Normative ID handling


james-ball-qualcomm (changes requested):
There's already a Makefile target in the ISA manual repo that extracts all the normative rule tags into JSON files. I'd much prefer you use that and consume the JSON files than extract the tags with yet another tool. Hopefully you just didn't know about this other tool.

Just run "make build-tags" and it will create JSON files for each ISA manual in the repos build directory. I also call this via UDB to generate the JSON files too and consume those in UDB.

PR #792, opened 2025-09-10
Made modifications to VM_PMP_cg

Zain2050:
We are supposed to test accesses by restricting pmp permissions of PTEs. I have corrected its coverpoints. We can't use PTE_d in these crosses because the PTE does not have read PMP permission, therefore we can't read it. We can check the progress of page table walk by the help of PageType_d. 
As we are testing for VM_PMP exceptions in this coverage file, therefore I am only testing for no read PMP permission, because during a page table walk the PTE is only read, not written or executed. So I don't think that we need to test it for W=0 & X=0.
Also, we don't get read/write/execute access signals during the page table walk. Therfore, I have replaced those with sw,lw & jalr.

### DIFF
--- a/fcov/rv64_priv/RV64VM_PMP_coverage.svh
+++ b/fcov/rv64_priv/RV64VM_PMP_coverage.svh
@@ -18,17 +18,20 @@
 // and limitations under the License.
 ////////////////////////////////////////////////////////////////////////////////////////////////
 
+`define sv48
+`define sv39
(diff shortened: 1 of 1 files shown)


davidharrishmc:
@UmerShahidengr could you review because you know these better?  When you are satisfied, we can merge.

UmerShahidengr (approved):
This looks good to me. We can merge this one

